PR #7, opened 2017-08-03
Add factory_girl

jcoyne:


app/models/source_file.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class SourceFile < ApplicationRecord
```
cbeer: This seems suspiciously similar to `Spotlight::Resource`, no?

PR #30, opened 2017-08-05
Use headless chrome for capybara feature tests

cbeer:


jcoyne:
We don't have any javascript tests, is this premature?

cbeer:
I guess it should read "feature tests".

jcoyne:
But the Rack driver is good enough for now, right?

cbeer:
Sure; I guess we can shelve this for now.

PR #38, opened 2017-08-07
Convert MODS to the IR

jcoyne:


Gemfile
```diff
@@ -86,9 +86,11 @@ gem 'devise-guests', '~> 0.6'
 
 gem 'config'
 
+gem 'github_api', github: 'piotrmurach/github'
```
cbeer: Why do we need this from github? Maybe leave a comment so we know when to move back to a release?

bin/fetch_and_import
```diff
@@ -0,0 +1,31 @@
...
+# This retrieves all the files in the directory corresponding to:
+# https://github.com/waynegraham/dlme-metadata/tree/master/maps/records/stanford
+# 1. Get a personal access token from GitHub (https://github.com/settings/tokens)
+#    with the following scopes enabled:
+#    * public_repo
+# 2. Set an ENV variable named 'GITHUB_TOKEN' containing your token
+
+AUTHORIZATION_TOKEN = ENV['GITHUB_TOKEN'] ||
+                      raise('GitHub authorization token was not found in the GITHUB_TOKEN environment variable')
+USER = 'waynegraham'
```
cbeer: I wonder if these should move into `config/settings.yml`?

PR #46, opened 2017-08-08
Copy index styling from exhibits to align thumbnails a little better

cbeer:
screen shot 2017-08-08 at 7 55 10 am


app/controllers/catalog_controller.rb
```diff
@@ -9,6 +9,7 @@ class CatalogController < ApplicationController
     config.show.oembed_field = :oembed_url_ssm
     config.show.partials.insert(1, :oembed)
 
+    config.view.list.partials = [:thumbnail, :index_header, :index]
```
jcoyne: Rubocop issue here.

PR #52, opened 2017-08-08
Improve full text searching for titles, contributors, etc

cbeer:


config/initializers/spotlight_initializer.rb
```diff
@@ -24,7 +24,7 @@
 # Spotlight::Engine.config.solr_batch_size = 20
 # Spotlight::Engine.config.filter_resources_by_exhibit = true
 # Spotlight::Engine.config.autocomplete_search_field = 'autocomplete'
-# Spotlight::Engine.config.default_autocomplete_params = { qf: 'id^1000 full_title_tesim^100 id_ng full_title_ng' }
+Spotlight::Engine.config.default_autocomplete_params = { qf: 'id^1000 full_title_tsim^100 id_ng full_title_ng' }
```
jcoyne: Should this be `cho_title_tsim`?

solr/config/schema.xml
```diff
@@ -15,7 +15,7 @@
     <field name="lng" type="tdouble" stored="true" indexed="true" multiValued="false"/>
     <field name="original_pid_tesim" type="pid_text" stored="true" indexed="true" multiValued="true"/>
 
-    <field name="full_title_ng" type="text_en_ng" stored="false" indexed="true" multiValued="true"/>
+    <field name="cho_title_ng" type="text_en_ng" stored="false" indexed="true" multiValued="true"/>
```
jcoyne: Some of our titles are non-english, this won't cause a problem, will it?

config/initializers/spotlight_initializer.rb
cbeer: thanks

solr/config/schema.xml
cbeer: I don't /think/ so. This is just an ngrams field to support autocomplete in sir trevor widgets, so a little slop here and there oughtn't make a big difference.

PR #65, opened 2017-08-10
MODS Mapping, Local Methods, Config, Tests

cmh2166:
- [x] update mappings doc for full MODS
- [x] add data provider / institution to Traject config
- [x] generate functions for MODS fields that require lookups beyond Xpath


cbeer:
👍 this seems like a clear improvement to our MODS indexing; how do you feel about getting it merged in early and often?

cmh2166:
alright, not able to get the local mappings i wanted that would require more work than i have time for right now, but i think everything we can guess at for mvp is in there. moving other bits to new branches / prs.

lib/traject/mods_config.rb
```diff
@@ -11,15 +12,102 @@
...
+# WR
+to_field 'wr_description', extract_mods('/*/mods:physicalDescription/mods:digitalOrigin')
+to_field 'wr_description', extract_mods('/*/mods:physicalDescription/mods:reformattingQuality')
+to_field 'wr_format', extract_mods('/*/mods:physicalDescription/mods:internetMediaType')
+
+# STANFORD Specific
+to_field 'wr_is_referenced_by', lambda { |_record, accumulator, context|
+  accumulator << if context.settings.fetch('identifier')
+                   inst = context.settings.fetch('inst_id') + '_'
+                   druid = context.settings.fetch('identifier').sub! inst, ''
```
cbeer: I think you want `#sub` (creates a new string) instead of `#sub!` (mutates the original) here.

cbeer:
Sounds good to me; just one small comment (and fixing the build).

docs/mappings.md
```diff
@@ -0,0 +1,139 @@
...
+mods:tableOfContents	|	cho_description	|
+mods:targetAudience	|	n/a	|
+mods:title/mods:nonSort	|	n/a	|
+mods:title/mods:partName	|	cho_title	|
+mods:title/mods:partName	|	n/a	|	Append to mods:titleInfo/mods:title[not(@\*)] cho_title?
+mods:title/mods:partNumber	|	cho_title	|
+mods:title/mods:partNumber	|	n/a	|	Append to mods:titleInfo/mods:title[not(@\*)] cho_title?
+mods:titleInfo/mods:subTitle	|	cho_title	|
+mods:titleInfo/mods:subTitle	|	n/a	|	Append to mods:titleInfo/mods:title[not(@\*)] cho_title?
+mods:titleInfo/mods:title[@type]	|	cho_alternative	|
```
cbeer: Should this be `mods:titleInfo[@type]`?

lib/traject/macros/stanford.rb
```diff
@@ -0,0 +1,96 @@
...
+    NS = { mods: 'http://www.loc.gov/mods/v3',
+           rdf: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#',
+           dc: 'http://purl.org/dc/elements/1.1/' }.freeze
+    # SOURCE SPECIFIC METHODS
+    # Macros for extracting Stanford-specific values from Nokogiri documents
+    def generate_sul_manifest
+      lambda { |record, accumulator, context|
+        druid = generate_druid(record, context)
+        # eventually add a test that calls IIIF service to check
+        if druid.present? && druid?(druid.chomp('druid:'))
```
cbeer: I wonder if should teach `druid?` how to handle the `druid:` prefix, and maybe also handle `nil` gracefully?

lib/traject/macros/stanford.rb
```diff
@@ -0,0 +1,96 @@
...
+          manifest = "https://purl.stanford.edu/#{druid}/iiif/manifest"
+          accumulator << manifest
+          grab_sul_iiif_links(manifest)
+          generate_sul_shown_at(druid)
+        end
+      }
+    end
+
+    def generate_druid(record, context)
```
cbeer: I think `generate_druid` and `druid?` (and probably some other internal methods) can/shuld be marked as private?

lib/traject/macros/stanford.rb
```diff
@@ -0,0 +1,96 @@
...
+    def generate_sul_shown_at(druid)
+      to_field 'agg_is_shown_at' do |record, accumulator|
+        mods_url = record.xpath('/*/mods:location/mods:url', NS).map(&:text)
+        accumulator.concat(mods_url) if mods_url.present?
+        accumulator << "https://purl.stanford.edu/#{druid}"
+      end
+    end
+
+    def grab_sul_iiif_links(manifest)
+      result = Net::HTTP.get(URI.parse(manifest))
```
cbeer: Instead of using `Net::HTTP` directly, I think we should explicitly add `faraday` as a dependency and use that. It's a little more flexible and handles some basic work for you... I think it'll also let us check whether the response was a success instead of throwing exceptions.

lib/traject/macros/stanford.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    def generate_sul_preview(thumbnail_uri)
+      to_field 'agg_preview', literal(thumbnail_uri)
+    end
+
+    def iiif_service(sequences)
+      to_field 'wr_has_service' do |accumulator|
+        sequences.each do |sequence|
+          sequence['canvases'].each do |canvas|
+            canvas['images'].each do |image|
```
cbeer: I think we decided yesterday with @anarchivist that the `agg_is_shown_by` only needs the thumbnail and/or first image of a series.

lib/traject/macros/mods.rb
```diff
@@ -16,9 +18,73 @@ def extract_name(role: nil, exclude: nil)
...
+
+        if url.present?
+          accumulator.concat(url)
+        elsif title.present?
+          accumulator.concat(title)
+        end
+      }
+    end
+
+    def generate_series
```
jcoyne: This code doesn't appear to be well tested.  Can you write a test for it?

lib/traject/macros/mods.rb
```diff
@@ -16,9 +18,73 @@ def extract_name(role: nil, exclude: nil)
...
+
+        if url.present?
+          accumulator.concat(url)
+        elsif title.present?
+          accumulator.concat(title)
+        end
+      }
+    end
+
+    def generate_relation
```
jcoyne: This method doesn't appear to be well tested. Can you write a test for it?

cmh2166:
Note, I'll rebase to clean up history when we're ready to go.

jcoyne:
@cmh2166 looks like the tests are very close... Just need a little more work.

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting Stanford Specific MODS values from Nokogiri documents
+  module Stanford
+    def grab_sul_iiif_links(manifest)
+      manifest = Net::HTTP.get(URI.parse(manifest))
+      JSON.parse(manifest)
+    rescue
+      nil
```
jcoyne: Because this could return nil, then we should test that the other methods that take this result as a parameter are able to work when you pass them `nil`, right?

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    def process_iiif_thumbnail(manifest_json)
+      manifest_json['thumbnail']['@id']
+    rescue JSON::ParserError
```
jcoyne: Seeing as the json is already passed in, is there some reason to expect a ParserError here? (same question applies to several methods below)

lib/traject/mods_config.rb
```diff
@@ -2,24 +2,159 @@
...
+to_field 'agg_is_shown_at' do |record, accumulator, context|
+  druid = generate_druid(record, context)
+
+  accumulator << transform_values(context, wr_id: literal(generate_sul_shown_at(record, druid)))
+end
+
+to_field 'agg_is_shown_by' do |record, accumulator, context|
+  druid = generate_druid(record, context)
+  manifest = "https://purl.stanford.edu/#{druid}/iiif/manifest"
+  iiif_json = grab_sul_iiif_links(manifest)
```
jcoyne: do we need to `continue/break/return` when `iiif_json` is nil?

lib/traject/macros/stanford.rb
cmh2166: done.

cmh2166: would be nice to log errors, handle better, but keeping simple just to get this through for right now.

lib/traject/macros/stanford.rb
cmh2166: yes, i think that's being produced now... i could add support for if no first image, use thumbnail, or vice versa, or something more. but right now it grabs the first sequence image details for the agg_is_shown_by, and the agg_preview grabs all the info for the thumbnail.

lib/traject/macros/mods.rb
cmh2166: Will do if settled.

lib/traject/macros/stanford.rb
anarchivist: Yep, and the application profile suggests that `agg_is_shown_by` should have a cardinality of 0..1.

lib/traject/mods_config.rb
cmh2166: sorry, just seeing this comment. i was hoping to handle the error in the IIIF methods (albeit they still have ParserError from a recent rewrite, which you clocked already as incorrect). We'll probably end up needing both checks.

cmh2166:
hooray, passes. going to add one lost check for bad IIIF response, rebase ugly history, then ask someone to review / merge.

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,80 @@
...
+      elsif service_profile.include? 'http://iiif.io/api/auth/'
+        'http://iiif.io/api/auth/'
+      elsif service_profile.include? 'http://iiif.io/api/presentation/'
+        'http://iiif.io/api/presentation/'
+      elsif service_profile.include? 'http://iiif.io/api/search/'
+        'http://iiif.io/api/search/'
+      end
+    end
+
+    def iiif_protocol(thumbnail_service_profile)
```
cbeer: I think `iiif_protocol` is unused?

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    def process_iiif_sequences_service_protocol(manifest_json)
+      manifest_json['sequences'].first['canvases'].first['images'].first['resource']['service']['profile']
+    rescue ::JSON::ParserError
+      nil
+    end
+
+    def process_iiif_sequences_conforms_to(manifest_json)
+      profile = manifest_json['sequences'].first['canvases'].first['images'].first['resource']['service']['profile']
```
cbeer: I wonder if we should extract `manifest_json['sequences'].first['canvases'].first['images'].first['resource']` to a method like:

```
def representative_iiif_resource(manifest_json)
  manifest_json['sequences'].first['canvases'].first['images'].first['resource']
end
```

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,80 @@
...
+    def grab_sul_iiif_links(manifest)
+      resp = Faraday.get URI.parse(manifest)
+      ::JSON.parse(resp.body)
+    rescue Faraday::Error
+      nil
+    end
+
+    def process_iiif_thumbnail(manifest_json)
+      manifest_json['thumbnail']['@id']
+    rescue ::JSON::ParserError
```
cbeer: I don't think any of these methods should raise `::JSON::ParserError`, do they? Does this belong on `grab_sul_iiif_links`?

lib/traject/macros/mods.rb
cbeer: Maybe we could pull out a simple method that all these could share? I'm not sure it's worth writing individual tests for each variation, but between all 3 we might cover them with our integration test?

lib/traject/macros/iiif.rb
cmh2166: yep. broken out. removing.

lib/traject/macros/iiif.rb
cmh2166: for sure, that's a nice way to simplify.

lib/traject/macros/iiif.rb
cmh2166: it was originally added to handle key doesn't exist errors. will see how to better handle and add.

cbeer: Got it. Here are some ideas:

```
manifest_json.dig('thumbnail', '@id')
```
(dig is new-ish, I've never used it, but this seems like a great spot for it, if it works)

The old style:

```
manifest_json.fetch('thumbnail', {}).fetch('@id')
```

or with lots of guard clauses:

```
manifest_json['thumbnail']['@id'] if manifest_json['thumbnail']
```

cmh2166: attempted a better repair for handling this.

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting Stanford Specific MODS values from Nokogiri documents
+  module IIIF
+    def grab_iiif_manifest(manifest)
+      resp = Faraday.get URI.parse(manifest)
+      ::JSON.parse(resp.body) if resp.status == 200
```
cbeer: I think `resp.success?` is the idiomatic version of `resp.status == 200` (and I think covers a couple other good-enough status codes?)

lib/traject/macros/iiif.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def iiif_sequence_service_protocol(iiif_json)
+      rep_iiif_resource(iiif_json).dig('service', 'profile') if rep_iiif_resource(iiif_json).present?
+    end
+
+    def iiif_sequence_service_conforms_to(iiif_json)
+      iiif_service_conforms_to(rep_iiif_resource(iiif_json).dig('service', 'profile')) if rep_iiif_resource(iiif_json).present?
+    end
+
+    def rep_iiif_resource(manifest_json)
```
cbeer: If you did something like:

```
manifest_json['sequences'].first['canvases'].first['images'].first['resource'] || {}
```

you wouldn't need to check `if rep_iiif_resource(iiif_json).present?` above, which might be nice.

lib/traject/macros/iiif.rb
cmh2166: sure.

lib/traject/macros/iiif.rb
cmh2166: sounds good.

PR #68, opened 2017-08-11
[WIP] Add language identification processing chain

cbeer:
This doesn't work so great with the documents we've indexed so far. The default certainty value (0.5) isn't low enough to produce any results, and lower values produce too much noise (particularly for text with lots of proper nouns, it seems).

### DIFF
--- a/solr/config/solrconfig.xml
+++ b/solr/config/solrconfig.xml
@@ -19,6 +19,9 @@
   <!-- solr lib dirs -->  
   <lib dir="${solr.install.dir:../../../..}/contrib/analysis-extras/lib" />
   <lib dir="${solr.install.dir:../../../..}/contrib/analysis-extras/lucene-libs" />
+  <lib dir="../../../dist/" regex="solr-langid-\d.*\.jar" />
+  <lib dir="${solr.install.dir:../../../..}/contrib/langid/lib" />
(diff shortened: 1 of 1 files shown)


cbeer:
Also, we either want https://issues.apache.org/jira/browse/SOLR-11231 / https://github.com/apache/lucene-solr/pull/232 fixed or need to set default values for fields we want to do langid on.

solr/config/solrconfig.xml
```diff
@@ -150,4 +153,21 @@
...
+    <lst name="defaults">
+      <str name="update.chain">langid</str>
+    </lst>
+  </initParams>
+
+  <updateRequestProcessorChain name="langid">
+    <processor name="langid" class="org.apache.solr.update.processor.TikaLanguageIdentifierUpdateProcessorFactory">
+      <lst name="defaults">
+        <str name="langid.fl">cho_title_tsim</str>
+        <str name="langid.langField">langid_ssim</str>
```
jcoyne: Are we using this field somewhere?

cbeer: Nope; once I saw the poor quality of the results, I stopped work on this.

justinlittman:
@cbeer Can this be closed?

jcoyne:
I'm going to make the call that we don't need this.

PR #70, opened 2017-08-11
Add CSV transform for traject

cbeer:


cmh2166:
Will look at this / test tomorrow (Sunday).

lib/traject/macros/csv.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting TEI values from Nokogiri documents
```
jcoyne: TEI?

lib/traject/macros/csv.rb
```diff
@@ -0,0 +1,15 @@
...
+  module Csv
+    # @param xpath [String] the xpath query expression
```
jcoyne: This is incorrect documentation

lib/traject/macros/csv.rb
```diff
@@ -0,0 +1,15 @@
...
+    def column(index_or_name, options = {})
+      lambda do |row, accumulator, _context|
+        result = Array(row[index_or_name])
+        result = result.map(&:strip) if options.key?(:trim)
```
jcoyne: Trimming is not tested. Should we remove this line?

cbeer: I think it's worth keeping, if only to ensure some consistency between our extraction options. I expect we'll soon be able to pull things like this and translation maps up into a more generic value filtering / mapping method?

PR #71, opened 2017-08-11
Add Traject JSON reader and IIIF configuration

cbeer:


cmh2166:
Do we have any IIIF metadata sources to load yet? Not seeing any in the metadata repo, and some iiif calls are dependent on adding source / institution logic (see the waiting PR #65, macros/stanford.rb).

Regardless, will review after getting through some other issues.

lib/traject/json_reader.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Reads in XML records for traject
```
jcoyne: XML?

PR #83, opened 2017-08-15
Don't calculate coverage on the spec directory

jcoyne:


spec/spec_helper.rb
```diff
@@ -27,6 +27,7 @@
   add_filter 'lib/traject/tei_config.rb'
```
cbeer: I think if we re-add `SimpleCov.start('rails') do` above, we'll get these normal filters for free, no?

jcoyne: @cbeer I tried that. When doing that, it disregards all the filters in the block.

jcoyne: See https://coveralls.io/builds/12862044

cbeer:
I think #87 is probably a better approach, and I'll continue to follow-up with the simplecov maintainers to see if there's something better we can do.

Issue #20, opened 2017-08-04
Develop mapping for Met Museum CSV


anarchivist:

cmh2166:
blocked by #18 (fgdc) - staging metadata mapping questions so it's less of a free for all resulting in lots of repetitive convos / decision processes.

PR #84, opened 2017-08-16
Met Museum CSV Mapping

aaron-collier:
This adds the csv mapping config for the Met Museum data. Including the traject macro for formatted fields (cho_creator, cho_date, cho_edm_type, cho_type, cho_dc_rights), a small edit to the fixture data so that the creator data was non-blank.

Addresses Issue #20 

This is a first pass, some DRY work and other cleanup is likely necessary for "production" - so I'm more than open to comment/suggestions.


lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting values from CSV rows
+  module MetCsv
+    # @param header_or_index [String] the field header or index to accumulate
+    def column(header_or_index, options = {})
```
cbeer: Can you include `Macros::Csv` and use the `column` definition from there instead of redefining it?

lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        if begin_date.present?
+          display_pattern = "#{display_pattern}, #{begin_date}"
+        end
+
+        if end_date.present?
+          display_pattern = "#{display_pattern} - #{end_date}"
+        end
+
+        if role.present?
```
cbeer: Do we need to worry about what happens if some of these values are missing or blank? It seems like the string formatting won't be quite right without it, yeah?

lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,95 @@
...
+      lambda do |row, accumulator, _context|
+        if row['Classification'].present?
+          accumulator << 'three dimensional object'
+        end
+      end
+    end
+
+    def public_domain
+      lambda do |row, accumulator, _context|
+        if row['Is Public Domain'].present? && row['Is Public Domain'] == 'True'
```
cbeer: Do you need the `present?` check?

lib/traject/macros/met_csv.rb
aaron-collier: For sure. On it.

lib/traject/macros/met_csv.rb
aaron-collier: Not likely here, was wondering about this. Will clean up.

lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,47 @@
...
+
+module Macros
+  # Macros for extracting values from CSV rows
+  module MetCsv
+    # @param header_or_index [String] the field header or index to accumulate
+
+    def generate_creator
+      lambda do |row, accumulator, _context|
+        display_pattern = row['Artist Display Name'] unless row['Artist Display Name'].empty?
+        display_pattern = "#{display_pattern}, #{row['Artist Suffix']}" unless row['Artist Suffix'].empty?
```
jcoyne: Can we use += operator here

jcoyne: What if display_pattern was nil here? Then it would begin with a comma and space. That seems odd

aaron-collier: Maybe - this can most certainly be better. But we're discussing a better structure for dealing with this, as well as reviewing what the pattern should look like when some of the data is blank. 

I'm not sure that should block this PR, and work through changes in the near future as we establish the creator pattern.

lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Macros
+  # Macros for extracting values from CSV rows
+  module MetCsv
+    # @param header_or_index [String] the field header or index to accumulate
+
+    def generate_creator
+      lambda do |row, accumulator, _context|
+        display_pattern = row['Artist Display Name'] unless row['Artist Display Name'].empty?
```
aaron-collier: Good call. Will update.

cmh2166:
Do we need both the csv_config.rb + the met_csv_config.rb? is one pulling from the other?

.rubocop_todo.yml
```diff
@@ -1,12 +1,20 @@
...
-# on 2017-08-01 11:56:46 -0700 using RuboCop version 0.49.1.
+# on 2017-08-16 16:23:48 -0700 using RuboCop version 0.49.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
 # Offense count: 1
+Metrics/AbcSize:
+  Max: 24
```
jcoyne: I'd prefer we correct this rather than upping the limit.  I believe this can be accomplished with a few of the suggestions below.

aaron-collier:
Constants added per @jcoyne request.

aaron-collier:
@cmh2166 - Good question. I started this as met_csv assuming that these are MET specific and could be different for another csv mapping. I'm happy to pull them into a single file if that's what makes the most sense, but seems separating them may set us up for future growth better.

lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,58 @@
...
+    ROLE = 'Artist Role'
+    BIO = 'Artist Display Bio'
+    CLASSIFICATION = 'Classification'
+    PUBLIC_DOMAIN = 'Is Public Domain'
+
+    def generate_creator
+      lambda do |row, accumulator, _context|
+        display_pattern = row[DISPLAY_NAME] unless row[DISPLAY_NAME].empty?
+        display_pattern = "#{display_pattern}, #{row[SUFFIX]}" unless row[SUFFIX].empty?
+        display_pattern = "#{display_pattern}, #{row[BEGIN_DATE]}" unless row[BEGIN_DATE].to_s.empty?
```
jcoyne: How about extracting `artist_date_range` method?  Additionally, extract `begin_date_present?` and `end_date_present?`

Can you do this? `display_pattern = [display_name, suffix, artist_date_range].select(&:present?).join(', ')`

lib/traject/macros/met_csv.rb
```diff
@@ -0,0 +1,58 @@
...
+        display_pattern = "#{display_pattern} - #{row[END_DATE]}" unless row[END_DATE].to_s.empty?
+        display_pattern = "#{display_pattern} (#{row[ROLE]}" unless row[ROLE].empty?
```
jcoyne: These two lines seem coupled (as the opening parenthesis is on one line and the closing parenthesis is on the other line).  Can this be extracted to a method?

lib/traject/macros/met_csv.rb
aaron-collier: This looks like a good option! Will pull it in.

lib/traject/macros/met_csv.rb
aaron-collier: For sure, as above. Will pull it in now.

lib/traject/macros/met_csv.rb
aaron-collier: Damn, this tip is a A LOT cleaner! Thanks @jcoyne!

lib/traject/met_csv_config.rb
```diff
@@ -0,0 +1,48 @@
...
+  provide 'reader_class_name', 'CsvReader'
+  provide 'allow_empty_fields', true
+end
+
+# MET Museum
+to_field 'id', column('Object ID')
+to_field 'cho_format', column('Object Name')
+to_field 'cho_creator', generate_creator
+to_field 'cho_spatial', column('City')
+to_field 'cho_has_type', column('Classification')
```
jcoyne: So we have `cho_has_type`, `cho_type` and `cho_edm_type` now?

aaron-collier: Yup - Review the mapping (https://docs.google.com/spreadsheets/d/1Sp7uMHizVX7xN7xN9mm-vgEuESQBovXO-qenAo_TV-w/edit#gid=0) lines 12, 47, 48

cmh2166: just capturing logic here + will make sure is clear in mapping docs too:
cho_format == form ; 
cho_has_type == genre ; 
cho_edm_type == normalized DCMI Type like ; 
cho_type == not normalized DCMI Type like. 
In the Met CSV, none of those are handled well (is mostly forms / genres in the Getty AAT vein), so mainly capturing the first two fields and inferring the last two.

I'm still trying to confirm the usage of 3D objects in places like MODS Types and relation to something like 3D objects in places like the EDM usage, and I think this may require a fix, but we can update that with a PR to not block this.

lib/traject/met_csv_config.rb
```diff
@@ -0,0 +1,48 @@
...
+to_field 'cho_spatial', column('County')
+to_field 'cho_provenance', column('Credit Line')
+to_field 'cho_coverage', column('Culture')
+to_field 'cho_provenance', column('Department')
+to_field 'agg_provider', column('Department')
+to_field 'cho_extent', column('Dimensions')
+to_field 'cho_coverage', column('Dynasty')
+to_field 'cho_coverage', column('Excavation')
+to_field 'cho_dc_rights', public_domain
+# to_field 'cho_source', column('Link Resource') # TODO: Blocked by PR: 65
```
cbeer: #65 is merged. We should update this mapping to add the structured information.

aaron-collier: I've added the `arg_is_shown_at` structured data.

cbeer: Should we remove this commented-out line? Is it still applicable?

aaron-collier: Removed. Sorry I missed that one.

PR #105, opened 2017-08-18
Increase the DB connection pool size for sidekiq

jcoyne:


config/initializers/sidekiq.rb
```diff
@@ -1,6 +1,13 @@
 if defined? Sidekiq
   Sidekiq.configure_server do |config|
     config.redis = { url: "redis://#{ENV['REDIS_HOST']}:#{ENV['REDIS_PORT']}/1" }
+
+    # we need to increase the db connection pool size to have enough connections
+    db_config = ActiveRecord::Base.configurations[Rails.env]
+    # the concurrency + 2 comes from: https://github.com/mperham/sidekiq/blob/master/4.0-Upgrade.md#whats-new
```
cbeer: Hm. I read that as about redis? Is the same true for the database?

jcoyne: Hmm, the +2 might be for redis.  But we do want at least one connection per sidekiq thread, right?

cbeer:
I think #106 should be, functionally, the same idea?

PR #108, opened 2017-08-18
Add a test for semantically invalid json

jcoyne:


jcoyne:
#109 should fix the problem with this build.

PR #113, opened 2017-08-18
Harvest remote resources into the database

jcoyne:


app/models/harvested_resource.rb
```diff
@@ -0,0 +1,28 @@
...
+  belongs_to :pipeline
+
+  def content
+    resource_content.body
+  end
+
+  # Return an identifier for use in the IR.  This is a hack that is needed
+  # because some of our metadata files don't have identifiers encodeded internally.
+  # Don't use this method if you have a better way of getting an identifier.
+  def identifier
```
cbeer: @aaron-collier and I were talking about maybe pushing this identifier logic down into the traject configs, and just making sure the original filename gets passed through (where it could then apply this logic). What do you think?

lib/github_importer.rb
```diff
@@ -7,20 +7,45 @@ def initialize(oauth_token, repo)
...
+    resource
+  end
+
+  # @param encoded [String] a Base64 encoded String
+  # @return [String] the decoded text with UTF-8 encoding
+  def decode(encoded)
+    Base64.decode64(encoded).force_encoding('UTF-8')
+  end
+
+  def sha1_to_multihash(sha1)
```
cbeer: What about pulling in https://github.com/multiformats/ruby-multihash? It seems pretty light-weight, at least. Either that, or can we include more information about multihashes are, where the `fn` comes from, etc.

app/models/harvested_resource.rb
jcoyne: That sounds good to me

lib/github_importer.rb
jcoyne: Sure, I added it.

app/models/harvested_resource.rb
jcoyne: https://github.com/sul-dlss/dlme/issues/117

cbeer:
Do you think the database will be ok with some of the large metadata formats we're getting / likely to get? Are we writing off things like carrierwave due to performance concerns?

jcoyne:
@cbeer what do you mean by large metadata?  I think TEXT is limited to 1GB.  Do we have to handle anything bigger that that?  I'm writing off carrierwave because I think sticking metadata in the database is the easiest and more platform independent way of doing this.

db/migrate/20170818154724_create_resource_contents.rb
```diff
@@ -0,0 +1,11 @@
+class CreateResourceContents < ActiveRecord::Migration[5.1]
+  def change
+    create_table :resource_contents do |t|
+      t.string :multihash, null: false
+      t.text :body, null: false
```
cbeer: Should we use `binary` here?

cbeer:
1 GB ought to be enough for anyone, huh? Can double check that size limit for sqlite and postgres? I have a dim memory of running into trouble with something like spotlight indexing logs that we were stashing in the database and maybe had to do something explicit to get the right size column.

db/migrate/20170818154724_create_resource_contents.rb
jcoyne: Aren't all our supported metadata formats text based?  Why would you consider binary? Are there any non-printing characters we mean to support?

jcoyne:
@cbeer 1GB was the limit I saw in PG.  

Sqlite says:
>  The maximum number of bytes in a string or BLOB in SQLite is defined by the preprocessor macro SQLITE_MAX_LENGTH. The default value of this macro is 1 billion (1 thousand million or 1,000,000,000). You can raise or lower this value at compile-time using a command-line option like this:
>
>    -DSQLITE_MAX_LENGTH=123456789
>
> The current implementation will only support a string or BLOB length up to 231-1 or 2147483647.

cbeer:
👌

PR #131, opened 2017-08-22
Use a branch of traject that allows Pathname configuration

jcoyne:


cbeer:
It looks like your upstream patch got merged; hopefully a release is imminent. If it isn't, couldn't we just call `#to_s` before passing them off to traject?

jcoyne:
@cbeer yes, that's a possible workaround

jcoyne:
This will be fixed in #147

PR #137, opened 2017-08-22
Deduplicate mods titles. Fixes #135

### LINKED ISSUE #135: Deduplicate values when transforming metadata into the IR
(Hopefully this is a traject configuration we can just enable?)

jcoyne:


cbeer:
Can we not use the traject `allow_duplicate_values` option?

jcoyne:
@cbeer I don't see how since we're doing multiple queries and concatenating all of them into one field.

jcoyne:
@cbeer Never mind I see now that that processing happens on the accumulator.

Issue #136, opened 2017-08-22
CSP is set incorrectly for GA and princeton IIIF


jcoyne:
```
Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src https: 'unsafe-eval' 'sha256-GFpwrduTw94NOGmBZ2ivuwcZIMsfyYQWMDV25hR82Ds=' https://www.google-analytics.com”). Source: onfocusin attribute on DIV element.  catalog
Content Security Policy: The page’s settings blocked the loading of a resource at http://libimages.princeton.edu/loris2/pudl0100%2Fposters%2Feg1_0095%2F00000001.jp2/full/200,/0/default.jpg (“img-src https://staging.dlme.clir.org https: data: https://www.google-analytics.com”).  (unknown)
Content Security Policy: The page’s settings blocked the loading of a resource at data:application/javascript;base64,KGZ1b... (“script-src https: 'unsafe-eval' 'sha256-GFpwrduTw94NOGmBZ2ivuwcZIMsfyYQWMDV25hR82Ds=' https://www.google-analytics.com”).  (unknown)
```

cbeer:
I'm not seeing the problem with GA in chrome. What browser / version are you using?

jcoyne:
This is Firefox 54

jcoyne:
I suspect this is because we're using jquery 1.x. https://bugzilla.mozilla.org/show_bug.cgi?id=1336076#c5  When I tried to remove it, I found we have 2 copies a jquery loaded via https://github.com/projectblacklight/spotlight/pull/1825

jcoyne:
The jquery update fixed several of the errors, now only one remains:

```
Content Security Policy: The page’s settings blocked the loading of a resource at data:application/javascript;base64,KGZ1b... (“script-src https: 'unsafe-eval' 'sha256-GFpwrduTw94NOGmBZ2ivuwcZIMsfyYQWMDV25hR82Ds=' https://www.google-analytics.com”).
```

jcoyne:
That last one comes from using uBlock on the page 😊

PR #139, opened 2017-08-22
Allow princeton to be an img_src. Ref #136

jcoyne:


cbeer:
Can we just "upgrade" their images to https as part of the transformation to the IR? It looks like it's a redirect from http -> https anyway.

jcoyne:
The princeton issue is only present in the fixture at this time. https://github.com/sul-dlss/dlme/blob/1db04e4b0bafc90822306f2d96e6d5820f26d96c/spec/fixtures/json/iiif-single-image.json#L19 Maybe we should just hold off on this for now?

cbeer:
Yeah; let's just update the fixture.

PR #147, opened 2017-08-23
Refactor import sources

### LINKED ISSUE #117: Move the logic that transforms filenames into IDs into the traject configs

cbeer:
Mainly, push the import configuration into our `settings.yml` so we can stop creating new classes and whatnot for each provider. I've proposed a provider configuration that looks like:

```yaml
    penn_tei:
      directory: manuscript/records/penn/schoenberg
      traject_file: tei_config
      properties:
        inst_id: penn
```

- `directory` (which is kinda a misnomer, since it's really a path in github)
- `traject_file` (though I'm open hear arguments for making this a full path / relative to the `Rails.root`)
- `properties` get passed through as traject context parameters.

Also fixes #117


app/jobs/github_import_job.rb
```diff
@@ -8,18 +8,18 @@ class GithubImportJob < ApplicationJob
   # @param harvest [Harvest] the harvest instance this job belongs to.
   # @param import_directory [String] the path in the github repository to import
   # @param pipeline_name [String] the name of the pipeline to use for transformation
-  def perform(harvest, import_directory, pipeline_name)
+  def perform(harvest, pipeline_name, config)
```
aaron-collier: Should the comments above be updated to reflect the change?

cbeer: 👍

app/jobs/github_import_job.rb
```diff
@@ -6,20 +6,20 @@
 # off the associated pipeline for each imported file
 class GithubImportJob < ApplicationJob
   # @param harvest [Harvest] the harvest instance this job belongs to.
-  # @param import_directory [String] the path in the github repository to import
   # @param pipeline_name [String] the name of the pipeline to use for transformation
-  def perform(harvest, import_directory, pipeline_name)
+  # @param config [Config::Options] configuration options for the pipeline
+  def perform(harvest, pipeline_name, config)
```
jcoyne: I'm wondering if the Pipeline could have the config?

PR #151, opened 2017-08-23
Allow the pipeline configuration to filter for specific files

cbeer:


config/settings.yml
```diff
@@ -29,4 +29,5 @@ import:
         inst_id: stanford
     met_csv:
       directory: museum/records
+      filter: !ruby/regexp '/met_museum_records.csv/'
```
jcoyne: Is this an exclusionary filter?  If so can we call it `excluding`?

cbeer: No, it's selecting resources that match.

jcoyne: Can we call it `only` then?

lib/github_importer.rb
```diff
@@ -12,6 +12,8 @@ def initialize(oauth_token, repo)
   # @yield [HarvestedResource] Gives the harvested resource
   def import(harvest, pipeline)
     gh.contents(repo, path: pipeline.config.directory).each do |resource|
+      next if pipeline.filter && resource.name !~ pipeline.filter
```
jcoyne: I don't understand how this is resources that match. It looks to me like we skip importing (`next`) if the file matches the filter. Right?

jcoyne: Doh. I just saw that you were using the not-matching operator.

cbeer: If there's a filter, and it doesn't match the `pipeline.filter`, we skip it.

config/settings.yml
cbeer: Sure.

lib/github_importer.rb
```diff
@@ -12,6 +12,8 @@ def initialize(oauth_token, repo)
   # @yield [HarvestedResource] Gives the harvested resource
   def import(harvest, pipeline)
     gh.contents(repo, path: pipeline.config.directory).each do |resource|
+      next if pipeline.filter && resource.name !~ pipeline.onlyu
```
jcoyne: typo

jcoyne: https://youtu.be/3FygIKsnkCw?t=10s

PR #152, opened 2017-08-23
updated types, downcase translation map, updated tests, closes #92

### LINKED ISSUE #92: review + repair 3D / Objects split among object types
Europeana to DCMI to MODS to MARC types mappings for `cho_type` and `cho_edm_type` aren't clear on handling objects (3d objects) vs digital representation types (i.e. 3d files).

Depending on decision, need to update translationMapping currently waiting in #65 and fix some config mappings.

cmh2166:


lib/traject/macros/xml.rb
```diff
@@ -21,6 +21,11 @@ def apply_translation(values, options)
       translation_map = Traject::TranslationMap.new(options[:translation_map])
       # without overwriting (further) translation map, could add
       # fuzzy match method here after pulling array out of TM
+      if values.is_a? Array
```
jcoyne: How about condensing this to `values = Array(values).map(&:downcase)`

lib/traject/macros/xml.rb
```diff
@@ -21,6 +21,11 @@ def apply_translation(values, options)
...
+        values = values.map(&:downcase)
+      elsif values.is_a? String
+        values = value.downcase
+      end
       translation_map.translate_array! values
```
cbeer: Can the value be a string? I thought all these traject things were arrays.

(Also, `translate_array!` below can/should probably be `translate_array`? )

lib/traject/macros/xml.rb
cmh2166: sure, sounds good

lib/traject/macros/xml.rb
cmh2166: some can be a string

lib/traject/macros/xml.rb
cmh2166: done.

PR #163, opened 2017-08-23
Move provider information into the configuration

cbeer:

PR #164, opened 2017-08-23
Fgdc mappings

### LINKED ISSUE #18: Develop mapping for Harvard FGDC

cmh2166:
Closes #18 .

### DIFF
--- a/lib/traject/fgdc_config.rb
+++ b/lib/traject/fgdc_config.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+require_relative 'xml_reader'
+require_relative 'dlme_json_resource_writer'
+require_relative 'macros/dlme'
+require_relative 'macros/fgdc'
+require_relative 'macros/xml'
+
+extend Macros::DLME
+extend Macros::Xml
+extend Macros::FGDC
+
+settings do
+  provide 'writer_class_name', 'DlmeJsonResourceWriter'
+  provide 'reader_class_name', 'XmlReader'
+end
+
+# Record Identifier
+to_field 'id', generate_fgdc_id
+
+# CHO Required
+to_field 'cho_identifier', generate_fgdc_id
+to_field 'cho_title', extract_fgdc('/*/idinfo/citation/citeinfo/title')
+to_field 'cho_title', extract_fgdc('/*/idinfo/citation/citeinfo/edition')
+
+# Other CHO
+to_field 'cho_coverage', extract_fgdc('/*/dataqual/lineage/srcinfo/srctime/timeinfo/rngdates/begdate')
+to_field 'cho_coverage', extract_fgdc('/*/dataqual/lineage/srcinfo/srctime/timeinfo/rngdates/enddate')
+to_field 'cho_coverage', extract_fgdc('/*/dataqual/lineage/srcinfo/srctime/timeinfo/sngdate/caldate')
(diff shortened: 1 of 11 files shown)


lib/traject/macros/xml.rb
```diff
@@ -17,6 +17,10 @@ def extract_xml(xpath, namespaces, options = {})
       end
     end
 
+    def extract_field(xml, xpath, namespaces)
```
cbeer: Is this method used?

cmh2166: is it expanded by the fgdc + mods xpath methods? i can check, but that's my understanding

lib/traject/fgdc_config.rb
```diff
@@ -0,0 +1,91 @@
...
+require_relative 'macros/xml'
+
+extend Macros::DLME
+extend Macros::Xml
+extend Macros::FGDC
+
+settings do
+  provide 'writer_class_name', 'DlmeJsonResourceWriter'
+  provide 'reader_class_name', 'XmlReader'
+  provide 'allow_empty_fields', true
```
jcoyne: We can remove this line.

lib/traject/fgdc_config.rb
```diff
@@ -0,0 +1,91 @@
...
+  provide 'agg_provider', 'Harvard University Library'
```
jcoyne: Move these three lines into `config/settings.yml` (You may need to rebase first)

lib/traject/macros/fgdc.rb
```diff
@@ -0,0 +1,56 @@
...
+    def harvard_uuid(record)
+      urls = record.xpath('/*/idinfo/citation/citeinfo/onlink', NS).map(&:text)
+      urls.first.split('CollName=')[1]
+    end
+
+    def extract_uuid(record, context)
+      institution = context.settings.fetch('agg_provider')
+      if institution.include?('Harvard')
+        harvard_uuid(record)
+      elsif institution.include?('MIT')
```
jcoyne: These xpaths looks the same to me. Can it be one elsif clause with `institution.include?('MIT') || institution.include?('Tufts')`

lib/traject/macros/xml.rb
jcoyne: Coveralls is showing no coverage of this method, so it's probably not used.

lib/traject/macros/fgdc.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        accumulator << identifier if identifier.present?
+      }
+    end
+
+    def select_identifier(record, context)
+      uuid = extract_uuid(record, context)
+      if uuid.present?
+        uuid
+      elsif context.settings.key?('command_line.filename')
```
jcoyne: In which cases are these `elsif` clauses called?  They aren't tested.  Are they supporting real use cases?

lib/traject/fgdc_config.rb
cbeer: (See #163 for more information about the config-driven approach)

cmh2166: yes, just seeing that. will do.

lib/traject/macros/fgdc.rb
cmh2166: done, thanks for that catch.

lib/traject/macros/fgdc.rb
cmh2166: the goal was to be a bit defensive if we ever get more than 4 fgdc records and we can't find ids. will add a test for it though.

PR #166, opened 2017-08-23
Adds Numismatic CSV mapping, test and fixture

aaron-collier:
This:
[x] Adds the numismatic csv traject config
[x] Adds a method to the csv macro to replace the "." to "_" in the numismatic id
[x] Adds a spec for the config
[x] Adds a one record fixture for the spec


lib/traject/numismatic_csv_config.rb
```diff
@@ -0,0 +1,49 @@
...
+require_relative 'dlme_json_resource_writer'
+require_relative 'macros/dlme'
+require_relative 'macros/csv'
+extend Macros::DLME
+extend Macros::Csv
+
+settings do
+  provide 'writer_class_name', 'DlmeJsonResourceWriter'
+  provide 'reader_class_name', 'CsvReader'
+  provide 'allow_empty_fields', true
```
cbeer: Can you remove the `allow_empty_fields` setting? I don't think we want/need it set.

lib/traject/numismatic_csv_config.rb
```diff
@@ -0,0 +1,49 @@
...
+  provide 'agg_provider', 'Numismatics Society'
```
cbeer: Instead of being added directly, the `agg_provider` data should probably be provided by the settings.yml

cbeer:
Can you add this new source to the `settings.yml`?

lib/traject/numismatic_csv_config.rb
cbeer: Sorry, this comment depends on https://github.com/sul-dlss/dlme/pull/163.

anarchivist: This also should be `'American Numismatic Society'`

lib/traject/numismatic_csv_config.rb
```diff
@@ -0,0 +1,49 @@
...
+end
+
+to_field 'agg_provider', provider
+
+# MET Museum
+to_field 'id', numismatic_id
+
+to_field 'cho_contributor', column('Authority')
```
anarchivist: I'm wondering if the pipes in `Authority` indicate multiple values. ANS's site is down so I can't confirm.

lib/traject/numismatic_csv_config.rb
anarchivist: Also, the value should be `'American Numismatic Society'`

config/settings.yml
```diff
@@ -31,3 +31,7 @@ import:
       directory: museum/records
       only: !ruby/regexp '/met_museum_records.csv/'
       traject_file: met_csv_config
+    numismatics:
+      directory: coins
+      traject_file: numismatic_csv_config
+      agg_provider: Numismatics Society
```
jcoyne: Change to "American Numismatic Society" per https://github.com/sul-dlss/dlme/pull/166#discussion_r134897279

lib/traject/macros/csv.rb
```diff
@@ -11,5 +11,11 @@ def column(header_or_index, options = {})
         accumulator.concat(result) unless result.first.to_s.empty?
       end
     end
+
+    def numismatic_id
+      lambda do |row, accumulator, _context|
+        accumulator << row['RecordId'].tr('.', '_')
```
jcoyne: Why are we changing dots to underscores?

cbeer: I believe it's about keep easily routable URL parameters (with the dots, we've have to go in and change the Rails default patterns.. which is fine and doable, but probably not worth it for this project).

Maybe it'd help if we called this method `normalize_numismatic_id`, and used, say, `row['RecordId'].to_param` to capture our intent a little better?

I also wonder if we want to use our institution-prefixed ids here (and anywhere else we're not)?

config/settings.yml
cbeer: Now that #163 is in, have a look how this was done on other providers - you need a little more hierarchy for these provided configurations.

lib/traject/macros/csv.rb
aaron-collier: That sounds good to me @cbeer, and yeah @jcoyne - this was because the dots are throwing routing errors.

spec/import/numismatic_csv_transform_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  it 'does the transform' do
+    indexer.process(data)
+    expect(CreateResourceJob).to have_received(:perform_later) do |_id, _two, json|
+      dlme = JSON.parse(json)
+      expect(dlme['agg_provider']).to eq 'American Numismatics Society'
+
+      # MET Museum
+      expect(dlme['id']).to eq 'ans_1975-93-862'
+
+      expect(dlme['cho_contributor']).to eq ['Zaffar b. Shabba|al-Qasim b. `Ubayd Allah']
```
anarchivist: In `Authority`, the pipe delimiter indicates multiple values (see http://numismatics.org/search/id/1975.93.862.xml).

spec/import/numismatic_csv_transform_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect(dlme['cho_medium']).to eq ['Glass']
+      expect(dlme['cho_creator']).to eq nil
+      expect(dlme['cho_format']).to eq ['Weight']
+      expect(dlme['cho_description']).to include '...sim b. ...Allah / Zaffar b....'
+      expect(dlme['cho_identifier']).to include '1975.93.862'
+      expect(dlme['cho_source']).to eq ['Morton.89']
+      expect(dlme['cho_spatial']).to eq ['NE Africa']
+      expect(dlme['cho_title']).to eq ['Glass 3 wuqiyya of Zaffar b. Shabba, NE Africa, 739 - 740. 1975.93.862']
+      expect(dlme['cho_identifier']).to include 'http://numismatics.org/search/id/1975.93.862'
+      expect(dlme['cho_date']).to eq ['739|740']
```
anarchivist: In `Year`, the pipe delimiter is for a range - see http://numismatics.org/search/id/1975.93.862.xml

jcoyne: 😱

lib/traject/macros/csv.rb
```diff
@@ -8,8 +8,22 @@ def column(header_or_index, options = {})
       lambda do |row, accumulator, _context|
         result = Array(row[header_or_index].to_s)
         result = result.map(&:strip) if options.key?(:trim)
+        result = (result.map { |s| s.split(options[:split]) }).first if options.key?(:split)
```
cbeer: Why `.first`?

lib/traject/macros/csv.rb
```diff
@@ -8,8 +8,22 @@ def column(header_or_index, options = {})
...
         accumulator.concat(result) unless result.first.to_s.empty?
```
cbeer: We can happily concat empty arrays all day, so this condition seems unnecessary?

lib/traject/macros/csv.rb
```diff
@@ -8,8 +8,22 @@ def column(header_or_index, options = {})
       lambda do |row, accumulator, _context|
         result = Array(row[header_or_index].to_s)
         result = result.map(&:strip) if options.key?(:trim)
```
cbeer: We should probably run `strip` after we do the splitting.

lib/traject/macros/csv.rb
aaron-collier: Otherwise the accumulator value becomes [['value 1', 'Value 2']] instead of ['Value 1', 'Value 2'].

aaron-collier: Wondering also how to bring this within the ABC issue...

lib/traject/macros/csv.rb
aaron-collier: Are you referring to the `if options` condition?

lib/traject/macros/csv.rb
aaron-collier: Will do.

lib/traject/macros/csv.rb
cbeer: Can we use `flat_map` or `flatten`?

cbeer: Or, maybe we should split the value earlier?

aaron-collier: flat_map looks like the key. thanks.

lib/traject/macros/csv.rb
cbeer: no, `unless result.first.to_s.empty?`

PR #167, opened 2017-08-23
	Map Princeton MODS metadata

### LINKED ISSUE #161: Implement Princeton MODS mapping

cbeer:
Fixes #161


spec/import/mods_transform_spec.rb
```diff
@@ -4,7 +4,13 @@
 
 RSpec.describe 'Transforming MODS files' do
   let(:provide) do
-    { 'identifier' => identifier, 'exhibit_slug' => slug, 'inst_id' => 'stanford' }
+    {
```
jcoyne: I don't think this block is used now that you've moved the definition to line 33.

lib/traject/macros/dlme.rb
```diff
@@ -14,6 +14,24 @@ def transform_values(context, hash)
       end
     end
 
+    # try a bunch of macros and short-circuit after one returns values
+    def first(*args)
```
jcoyne: Can we call `args` something more descriptive? How about `macros`?

cbeer: 👌

spec/import/mods_transform_spec.rb
cbeer: You're right; I've switched things around a little bit now so it does get used. Keeping it around will be useful when we add additional MODS mappings.

PR #172, opened 2017-08-24
Allow harvest & transform to be started in the UI.

jcoyne:


cbeer:
Looks like a good first stab to me, 👏

PR #186, opened 2017-08-25
Support dates that are just a year

jcoyne:
Fixes https://app.honeybadger.io/projects/53082/faults/34481413


cbeer:
I wonder if we should just leave everything as strings instead of applying the converter here:

https://github.com/sul-dlss/dlme/blob/master/lib/traject/csv_reader.rb#L12

Off-hand, I don't remember a good reason for doing so?

jcoyne:
@cbeer I removed the converter.

PR #187, opened 2017-08-25
Render thumbnails where solr has multiple values

jcoyne:
Fixes https://app.honeybadger.io/projects/53082/faults/34481424


cbeer:
Huh. I thought our IR had those as single-valued. 🤷‍♂️

jcoyne:
@cbeer they are single in the IR, but this is in Solr, where they are multivalued.

PR #189, opened 2017-08-25
Don't add extra characters to the creator. Fixes #183

### LINKED ISSUE #183: Met mapping - `cho_creator` has extraneous characters
It appears that the `artist_role_bio` method in the Met macros is adding extraneous characters .

* If the `Artist Display Name` is column is empty for a given row in the MetCSV, `cho_creator` should be blank.
* If `Artist Role` and `Artist Display Bio` is blank, do not include parentheses.

screen shot 2017-08-24 at 3 55 24 pm

jcoyne:


lib/traject/macros/met_csv.rb
```diff
@@ -40,11 +40,14 @@ def generate_creator
...
 
     def artist_date_range(row)
-      @artist_date_range ||= [row[BEGIN_DATE], row[END_DATE]].select(&:present?).join(' - ')
+      [row[BEGIN_DATE], row[END_DATE]].select(&:present?).join(' - ')
     end
 
     def artist_role_bio(row)
-      @artist_role_bio ||= "#{artist_date_range(row)} (#{[row[ROLE], row[BIO]].select(&:present?).join(' ; ')})"
+      role = [row[ROLE], row[BIO]].select(&:present?).join(' ; ')
+      artist = artist_date_range(row)
```
cbeer: is `artist` the right variable name? it seems like a date to me.

jcoyne: On this line it is a date range, on the next line I add the role & bio to this variable.

cbeer: Sure, but it's never an artist name, is it? `artist_info`, maybe?

jcoyne: Sure, names are hard.

PR #191, opened 2017-08-25
Change the join logic for display fields

### LINKED ISSUE #182: Changing joining for multi-valued fields in fielded display for on index and show pages
* Multiple values for `cho_provenance`and `cho_description` should be broken out into new paragraphs 
* Multiple values for all other fields should be broken into new lines

screen shot 2017-08-24 at 3 47 36 pm
screen shot 2017-08-24 at 3 47 11 pm

jcoyne:
Join description and provinance with paragraphs and other fields with
linebreaks.

Fixes #182

### DIFF
--- a/app/processors/join.rb
+++ b/app/processors/join.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+# Joins values using configured value or linebreak
+class Join < Blacklight::Rendering::AbstractStep
+  include ActionView::Helpers::TextHelper
+
+  def render
+    # rubocop:disable Rails/OutputSafety
+    joiner = config.join_with || '<br>'.html_safe
+    # rubocop:enable Rails/OutputSafety
+    next_step(safe_join(values, joiner))
+  end
+end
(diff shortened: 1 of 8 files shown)


app/processors/join.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Joins values using configured value or linebreak
+class Join < Blacklight::Rendering::AbstractStep
+  include ActionView::Helpers::TextHelper
+
+  def render
+    # rubocop:disable Rails/OutputSafety
+    joiner = config.join_with || '<br>'.html_safe
```
cbeer: I think this can be done now using `separator_options` (used as arguments to https://apidock.com/rails/Array/to_sentence)? It's not quite as clean, since we'll probably have to set all 3 options, but I think it should work.

jcoyne: @cbeer Yes, I considered that, but then we'd have to set separator_options on every field.  This seems nicer.

cbeer: :shrug:. I guess we can revisit it if we ever want to use the upstream code too.

PR #197, opened 2017-08-25
Add Penn Egyptian Museum CSV mapping

aaron-collier:
This adds the penn mapping and updates the met museum id.

Note: there is DRY here that should be reworked in a follow up PR, but we want this work in now to prep the ID especially for met.


lib/traject/macros/csv.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# require_relative 'tranforms'
+# extend Macros::Tranforms
```
jcoyne: Remove?

lib/traject/macros/met_csv.rb
```diff
@@ -25,7 +25,7 @@ def met_thumbnail
 
         return if image_json.blank?
         # Some records e.g. 321624, don't return any results
-        result = image_json['results'].first
+        result = image_json['results'].first unless image_json['results'].nil?
```
jcoyne: How about `Array(image_json['results']).first` then there's no conditional.

spec/import/met_csv_transform_spec.rb
```diff
@@ -2,37 +2,30 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Transforming MET CSV files' do
+RSpec.describe 'TransformingMetropolitan Museum of Art CSV file' do
```
jcoyne: missing a space here.

spec/import/met_csv_transform_spec.rb
```diff
@@ -2,37 +2,30 @@
...
   before do
-    allow(Traject::JsonWriter).to receive(:new).and_return(writer)
+    allow(CreateResourceJob).to receive(:perform_later)
   end
 
   it 'does the transform' do
     indexer.process(data)
-    expect(writer).to have_received(:put) do |context|
-      dlme = context.output_hash
-      expect(dlme['agg_preview'])
```
jcoyne: why did you drop this test?

aaron-collier:
@jcoyne I believe your requests have been addressed.

lib/traject/macros/met_csv.rb
```diff
@@ -25,7 +25,7 @@ def met_thumbnail
 
         return if image_json.blank?
         # Some records e.g. 321624, don't return any results
-        result = image_json['results'].first
+        result = Array(image_json['results']).first
```
jcoyne: Can you explain where you are seeing the json that doesn't have a `results` entry?

spec/fixtures/csv/met.csv
```diff
@@ -1,8 +1,4 @@
 "Object Number","Is Highlight","Is Public Domain","Object ID","Department","Object Name","Title","Culture","Period","Dynasty","Reign","Portfolio","Artist Role","Artist Prefix","Artist Display Name","Artist Display Bio","Artist Suffix","Artist Alpha Sort","Artist Nationality","Artist Begin Date","Artist End Date","Object Date","Object Begin Date","Object End Date","Medium","Dimensions","Credit Line","Geography Type","City","State","County","Country","Region","Subregion","Locale","Locus","Excavation","River","Classification","Rights and Reproduction","Link Resource","Metadata Date","Repository"
-"74.51.4366","False","True",321381,"Ancient Near Eastern Art","Stamp seal","Quasi-pyramidal seal","Cypriot","Late Cypriot III","","","","Demo Role","Demo Prefix","Demo Display Name","Demo Artist Bio","Demo Suffix","","","1900","2000","ca. 12th�11th century B.C.",-1200,-1000,"Steatite, gray brown","Seal Face: 1.19 x 1.68 cm
```
jcoyne: Why are these rows removed?

lib/traject/macros/met_csv.rb
aaron-collier: I was getting a syntax error here when the array was nil?... I'll have to dive back into that.

spec/fixtures/csv/met.csv
aaron-collier: The spec was failing on the ID check with multiple lines. It seemed to grab a random one of the 3 that were in there. This may have been related to the wonky spec difference I had. I'll restore the fixture and try it again.

jcoyne: Yeah, it's because there are carriage returns in the table cells. That's why I grabbed the first four lines, because that's equivalent to the headers + 1 record.

lib/traject/macros/met_csv.rb
jcoyne: Can you revert that and see how you got that error?

aaron-collier: Yeah. Working on it now.

jcoyne:
Please remove `.DS_Store`. Consider making a global gitignore https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #198, opened 2017-08-25
Add footer content partial

### LINKED ISSUE #118: Footer Content
We need an are to include content for the site-wide footer. Maybe two columns, with `col-md-2` and `col-md-10` for a **Contact** column and **About** section.

waynegraham:
Adds a footer partial (with content) for the site.

Closes #118

### DIFF
--- a/app/assets/stylesheets/footer.scss
+++ b/app/assets/stylesheets/footer.scss
@@ -1,0 +1,36 @@
+$clir-burgundy: #560319;
+$dlf-blue:      #046BA5;
+$bg-gray:       #eef2f5;
+
+
+@import 'https://fonts.googleapis.com/css?family=Crimson+Text:600';
+@import 'https://cdn.rawgit.com/clirdlf/logo-fonts/master/style.min.css';
+@import 'https://cdn.rawgit.com/clirdlf/logo-fonts/master/clir-font/stylesheet.min.css';
+@import 'https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css';
+
+.bg-gray {
+  background: $bg-gray;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/footer.scss
```diff
@@ -0,0 +1,36 @@
+$clir-burgundy: #560319;
+$dlf-blue:      #046BA5;
+$bg-gray:       #eef2f5;
+
+
+@import 'https://fonts.googleapis.com/css?family=Crimson+Text:600';
+@import 'https://cdn.rawgit.com/clirdlf/logo-fonts/master/style.min.css';
+@import 'https://cdn.rawgit.com/clirdlf/logo-fonts/master/clir-font/stylesheet.min.css';
+@import 'https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css';
```
jcoyne: This is going to add like about 200k to the page weight just for 3 icons? Are you sure?

jcoyne: Do these domains need to be added to the CSP headers in `config/initializers/secureheaders.rb`?

jcoyne: By making these `@import` aren't we loosing any benefit that a cdn would bring?

cbeer: I think we should make a note to revisit these questions later. I certainly share your concern about these being a little heavy-handed for 3 social media icons and 2 logos, but 🤷‍♂️

waynegraham:
@ggeisler may have some suggestions on this too

cbeer:
Great. I got it merged in in #212 and we can work to refine it over the week.

PR #204, opened 2017-08-25
Enable blacklight's facet 'more >' link

cbeer:


cmh2166:
works great for me locally. seems easy enough to revert if people decide they don't like this facet limit.

PR #207, opened 2017-08-26
Add utility method for fetching and parsing JSON from an external service

cbeer:


cmh2166:
Are we waiting for other tests here? It seems like coveralls is just failing bc of minimal decrease in coverage for test coverage on parts outside of scope of this PR. I'm inclined to merge this and make tickets for things lacking coverage.

cbeer:
Yeah, as you observed, the coverage drop doesn't have anything to do with this PR, so I'm 👍 to shipping.

PR #209, opened 2017-08-26
Pull out a TransformPipeline class with steps for every option

cbeer:
Includes #208


cmh2166:
I believe this changes the ability to run Traject locally without then also doing indexing / loading / i.e. the rest of the transform pipeline. Can someone else verify?

cmh2166:
this seems to be only me. will merge after lunch.

PR #210, opened 2017-08-26
Wip extraction pattern

cbeer:
🤷‍♂️


justinlittman:
@cbeer Can this be closed?

jcoyne:
This was all moved into TrajectPlus, so I don't think it has any remaining value.

PR #214, opened 2017-08-28
Add Penn Near East Config

aaron-collier:
Adds the Penn near east configuration, spec and fixture.


lib/traject/penn_near_east_config.rb
```diff
@@ -0,0 +1,55 @@
...
+to_field 'agg_provider', provider
+
+# MET Museum
+to_field 'id', normalize_prefixed_id('emuIRN')
+
+to_field 'cho_provenance', column('accession_credit_line')
+to_field 'cho_creator', column('creator')
+to_field 'cho_coverage', column('culture', split: '|')
+to_field 'cho_coverage', column('culture_area')
+to_field 'agg_data_provider', normalize_penn_egyptian_provider
```
cbeer: Do you want to rename this macro now that it's used in near east too?

lib/traject/penn_near_east_config.rb
```diff
@@ -0,0 +1,55 @@
...
+to_field 'cho_date', column('date_made_early')
+to_field 'cho_date', column('date_made_late')
+to_field 'cho_description', column('description')
+to_field 'cho_identifier', column('emuIRN')
+to_field 'cho_subject', column('iconography')
+to_field 'cho_medium', column('material', split: '|')
+to_field 'cho_extent', column('measurement_height')
+to_field 'cho_extent', column('measurement_length')
+to_field 'cho_extent', column('measurement_outside_diameter')
+to_field 'cho_extent', column('measurement_tickness')
```
cbeer: I assume `tickness` is how it is in the data? Should we have a ticket to follow up with the data provider?

lib/traject/penn_near_east_config.rb
```diff
@@ -0,0 +1,55 @@
...
+to_field 'cho_extent', column('measurement_unit')
+to_field 'cho_extent', column('measurement_width')
+to_field 'cho_alternative_title', column('native_name')
+to_field 'cho_title', column('object_name', split: '|')
+to_field 'cho_source', column('object_number')
+to_field 'cho_identifier', column('other_numbers', split: '|')
```
cbeer: I'm confused about the format of this field. it looks comma-separated, maybe?

lib/traject/penn_near_east_config.rb
```diff
@@ -0,0 +1,55 @@
...
+
+to_field 'cho_provenance', column('accession_credit_line')
+to_field 'cho_creator', column('creator')
+to_field 'cho_coverage', column('culture', split: '|')
+to_field 'cho_coverage', column('culture_area')
+to_field 'agg_data_provider', normalize_penn_egyptian_provider
+to_field 'cho_date', column('date_made')
+to_field 'cho_date', column('date_made_early')
+to_field 'cho_date', column('date_made_late')
+to_field 'cho_description', column('description')
```
cbeer: If I got my field right, this one looks like it might be `<br />` separated?

lib/traject/penn_near_east_config.rb
aaron-collier: I copy and pasted from the spreadsheet. I'll check the data.

lib/traject/penn_near_east_config.rb
aaron-collier: We can. I'm actually refactoring this in a separate branch to keep it simple, but a rename may be in order for now.

lib/traject/penn_near_east_config.rb
aaron-collier: Let me verify, I'm pretty sure I found '|' in the data.

lib/traject/penn_near_east_config.rb
aaron-collier: I'll review.

lib/traject/penn_near_east_config.rb
aaron-collier: Done in next update.

lib/traject/penn_near_east_config.rb
aaron-collier: This is in fact in the data. Issue opened.

lib/traject/penn_near_east_config.rb
aaron-collier: The `other_numbers` column def has data like `I.104 a-b|B80 \| 6/22/80 \| I.1.91`

lib/traject/penn_near_east_config.rb
aaron-collier: It looks less `separated` and more `formatted`...

aaron-collier: Adding separation

PR #252, opened 2017-08-30
Restructure footer and remove its about content

ggeisler:
This PR assumes it is okay to move the previous About paragraph out of the footer because it will be represented in an about page. Also replaces social media icons with individual images (7KB total) rather than using Font Awesome. And makes the footer sticky to bottom of viewport.

@waynegraham Please check screenshots below and make sure you are okay with the content updates to the footer.

The gray background of the footer will look a bit better with a forthcoming PR addressing other visual design updates.

### Updated page with restructured footer
site-footer

### On mobile
site-footer-mobile


waynegraham:
Looks good to me!

Issue #236, opened 2017-08-30
More facet does not expand.


jcoyne:
There are two problems here caused by  Blacklight's compatibility with jQuery 3.   The first and easiest to fix is that we need to use `jQuery.length` rather than `jquery.size()`.

The second problem is that calling `$.ajax({dataType: 'script'})` evaluates the returned value as javascript in jquery 3.  However our rails app is returning a HTML partial. This causes a syntax error.

```
$.ajax({
    url: $(this).attr('href'),
    dataType: 'script'
  });

```

Ref https://github.com/projectblacklight/blacklight/issues/1759

PR #253, opened 2017-08-30
Point at a branch of blacklight that supports jquery3. Ref #236

jcoyne:


Gemfile
```diff
@@ -67,6 +67,7 @@ end
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
+gem 'blacklight', github: 'projectblacklight/blacklight', branch: 'release-6.x'
```
cbeer: Can we just cut a new release?

jcoyne: Sure, are we done yet though?

cbeer: Done enough, no?

jcoyne: What about #222?

PR #254, opened 2017-08-30
Update visual design

ggeisler:
This PR should NOT be merged until after #252 (I based this branch on that branch so I could work with the new footer color and am worried about conflicts or other weirdness if we merge this branch first).

This PR updates various colors to better match the site header and footer. Also includes some other styling refinements to go with the visual theme. I only have one item in my local environment so there might be some minor additional styling I need to do when I see the results pages with multiple items.

I don't like that the SCSS variables I created are in specific SCSS files, rather than a global variables file, but I couldn't get a standalone `variables.scss` file to work, even though I've done it before on other projects. If someone knows what I'm missing we could refactor in a separate follow-on PR to clean things up.

### Example page 1

-_digital_library_of_the_middle_east_-_digital_library_of_the_middle_east_search_results

### Example page 2
about_dlme___digital_library_of_the_middle_east_-_digital_library_of_the_middle_east

### Example page 3
curation_-_dashboard___digital_library_of_the_middle_east_-_digital_library_of_the_middle_east


ggeisler:
If we merge #252 and we like this, is there a way to remove the "Restructure footer and remove its about content" commit from this PR before merging it? That commit is already in #252. 

I realize now creating the branch is this PR from the working branch in #252 just made things confusing. But maybe git knows how to deal with it all when merging?

PR #255, opened 2017-08-31
Update Manage Data page and superadmin sidebar

ggeisler:
Very minor updates:

- Change label for sidebar menu item to "Manage data" from "Harvests" so label is consistent with other sidebar actions (label starts with a verb and matches primary title of page is goes to).

- Change label for table caption on the Harvests page (because the table includes more info than just number of files) and darken text (because it'll need more contrast with the visual design updates in #254).

manage_data_-_harvests___digital_library_of_the_middle_east

### DIFF
--- a/app/assets/stylesheets/variables.scss
+++ b/app/assets/stylesheets/variables.scss
@@ -7,3 +7,5 @@ $color-gray-lightest: #f5f5f5;
 
 $clir-burgundy: #560319;
 $dlf-blue:      #046BA5;
+
+$text-muted: $color-black-medium; // with gray body bg, need darker muted color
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/dlme.scss
```diff
@@ -19,3 +19,7 @@
   width: 80%;
   margin-left: 10%;
 }
+
+caption {
+  color: #333;
```
cbeer: We probably want to change `text-mute` globally, yeah? And, let's figure out why a stand-alone variables file isn't working, if that's the blocker.

ggeisler:
Yep, will do.

PR #261, opened 2017-08-31
Record additional context within the DlmeJson and index it appropriately

cbeer:
For now, this context may include these settings:
 - command_line.filename (the original provided name when running traject)
 - source (used as a key for our import configuration)
 - position (record N of M in a batch)
 - harvest_id
 - harvest_resource_id

### DIFF
--- a/app/models/pipeline.rb
+++ b/app/models/pipeline.rb
@@ -20,6 +20,8 @@ class Pipeline < ApplicationRecord
   private
 
   def traject_config(resource)
-    config.properties.to_h.merge('command_line.filename' => resource.original_filename)
+    config.properties.to_h.merge('command_line.filename' => resource.original_filename,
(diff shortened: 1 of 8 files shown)


app/models/pipeline.rb
```diff
@@ -20,6 +20,8 @@ def indexer(resource)
   private
 
   def traject_config(resource)
-    config.properties.to_h.merge('command_line.filename' => resource.original_filename)
+    config.properties.to_h.merge('command_line.filename' => resource.original_filename,
+                                 'harvested_resource_id' => resource.id,
+                                 'harvest_id' => resource.harvest&.id)
```
jcoyne: `resource.harvest_id`

cbeer: 👍

cbeer:
I'm going to add these as facets before we merge

Issue #238, opened 2017-08-30
Index the pipeline name and make it available as a facet and display field for administrators 


cbeer:
We think this would be useful to help debug mappings

cbeer:
I guess this is suspiciously similar to #242, oops.

PR #262, opened 2017-08-31
Add some administrative / data qa facets

cbeer:


app/controllers/catalog_controller.rb
```diff
@@ -67,6 +67,26 @@ class CatalogController < ApplicationController
     config.add_facet_field 'dc_rights',   field: 'cho_dc_rights_ssim', limit: true
     config.add_facet_field 'provided_by', field: 'agg_data_provider_ssim', limit: true
 
+    # "administrative"-like facets
```
jcoyne: see #238. Do we want these to be `if: :superadmin?` too?

cbeer: Eventually? I figured we could start here and see if they're useful first.

app/controllers/catalog_controller.rb
```diff
@@ -67,6 +67,26 @@ class CatalogController < ApplicationController
...
+    config.add_facet_field 'data provider', field: 'agg_provider_ssim', limit: true
```
anarchivist: We should find another label here, because this could be confusing (compare `agg_data_provider`, which drives the "Provided by" facet).

cbeer: can you suggest one? these are also changeable in the spotlight admin interface, so we don't need to perfect it now either.

PR #263, opened 2017-08-31
Facet on the pipeline for admins. Fixes #238

### LINKED ISSUE #238: Index the pipeline name and make it available as a facet and display field for administrators

jcoyne:


cbeer:
I have an alternative approach in #261. Maybe we can discuss at standup?

PR #273, opened 2017-09-01
Visual design tweaks

ggeisler:
A couple of tweaks to try to improve the issue, mostly only noticeable on search results pages, where the gray backgrounds of various elements visually clash with the background pattern of the page. I originally used this body background pattern to go better with the footer color, but I think the background color works well enough.

- Replace body background pattern with a (similar in tone) background color
- Remove thumbnail background color and border

creator__jurjani__isma il_ibn_ asan__1042_or_3-1136_or_7_-_digital_library_of_the_middle_east_-_digital_library_of_the_middle_east_search_results


app/assets/stylesheets/dlme.scss
```diff
@@ -21,7 +22,7 @@
 }
 
 body {
-  background-image: asset-url("dlme-background.png");
```
cbeer: Should we remove the file as well?

ggeisler: Yep, thanks for catching that. Just updated the PR.

app/assets/stylesheets/dlme.scss
```diff
@@ -150,8 +152,15 @@ body {
   border-top-color: $color-gray-medium;
 }
 
-.thumbnail {
-  background-color: $color-gray-light;
+.pagination>.active>a,
```
cbeer: Can these be set using the `$pagination-*` variables?

https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L455

ggeisler:
Made one more update to the PR; I noticed the default pagination color and changed it to a gray to go better with the visual theme.

app/assets/stylesheets/dlme.scss
ggeisler: Yeah, probably. I can't easily verify that approach works locally, since I only have one record and don't get the pagination controls, but using those variables should work (forgot they existed) so I'll redo and push up another version.

app/assets/stylesheets/variables.scss
```diff
@@ -13,3 +13,7 @@ $footer-top-margin: 20px;
 $footer-mobile-adjustment: 40px;
 
 $text-muted: $color-black-medium; // with gray body bg, need darker muted color
+
+$pagination-color: $color-blackish;
```
cbeer: I thought you were only applying the custom styles to the active page before?

ggeisler: You're correct. I hastily copied the wrong group of variables. Hopefully this last update is good.

ggeisler:
Phew! That was harder than it should've been. Thanks @cbeer.

PR #288, opened 2017-09-02
Feature/yale medical manuscripts

waynegraham:
Adds the Yale Medical Manuscript collection


lib/traject/yale_csv_config.rb
```diff
@@ -0,0 +1,46 @@
...
+to_field 'cho_is_part_of',   column('Yale Collection', trim: true)
+to_field 'wr_dc_rights',     column('Rights', trim: true)
+to_field 'cho_provenance',   column('part of, group', trim: true)
+to_field 'agg_is_shown_at' do |_record, accumulator, context|
+  accumulator << transform_values(
+    context,
+    'wr_id' => [column('Handle', trim: true)]
+  )
+end
+to_field 'agg_is_shown_by' do |_record, accumulator, context|
```
cbeer: I /think/ this is supposed to be a link to the digital representation of the object (like a high-res photo or something). If there isn't one available, I think we should omit it. @anarchivist can probably confirm?

lib/traject/yale_csv_config.rb
```diff
@@ -0,0 +1,46 @@
...
+
+to_field 'agg_provider',     provider
+to_field 'id',               normalize_prefixed_id('OrbisBibID')
+to_field 'cho_identifier',   column('OrbisBibID', trim: true)
+to_field 'cho_provenance',   column('Place of origin', trim: true)
+to_field 'cho_creator',      column('Creator', split: '|', trim: true)
+to_field 'cho_title',        column('Title', split: '|', trim: true)
+to_field 'cho_alternative',  column('Variant Titles', split: '|', trim: true)
+to_field 'cho_publisher',    column('Publisher', trim: true)
+to_field 'edm:dateProvider', column('Date, created', trim: true)
```
cbeer: I don't think this is the right IR field.

lib/traject/yale_csv_config.rb
anarchivist: Yes, that's correct. It should be a link to an asset itself, like a JPEG, TIFF, PDF, MP3, etc., not a link to an object in a digital library interface.

lib/traject/yale_csv_config.rb
anarchivist: This should probably be `cho_date`?

lib/traject/yale_csv_config.rb
```diff
@@ -0,0 +1,46 @@
...
+to_field 'cho_description',  column('Physical description', trim: true)
+to_field 'cho_language',     column('Language', trim: true)
+to_field 'cho_description',  column('Notes', split: '|', trim: true)
+to_field 'cho_description',  column('Abstract', trim: true)
+to_field 'cho_provenance',   column('Associated Names', trim: true)
+to_field 'cho_subject',      column('Subject, topic', split: '|', trim: true)
+to_field 'cho_edm_type',     column('Type of resource', trim: true)
+to_field 'cho_is_part_of',   column('Yale Collection', trim: true)
+to_field 'wr_dc_rights',     column('Rights', trim: true)
+to_field 'cho_provenance',   column('part of, group', trim: true)
```
anarchivist: I'd ask Yale about this - from the look of the data, I'm wondering if this is info about the grant that funded digitization.

anarchivist: Well, these don't seem to be digitized, so maybe I'm somewhat wrong... But still, I'd verify this mapping.

lib/traject/yale_csv_config.rb
```diff
@@ -0,0 +1,46 @@
...
+to_field 'edm:dateProvider', column('Date, created', trim: true)
+to_field 'cho_description',  column('Physical description', trim: true)
+to_field 'cho_language',     column('Language', trim: true)
+to_field 'cho_description',  column('Notes', split: '|', trim: true)
+to_field 'cho_description',  column('Abstract', trim: true)
+to_field 'cho_provenance',   column('Associated Names', trim: true)
+to_field 'cho_subject',      column('Subject, topic', split: '|', trim: true)
+to_field 'cho_edm_type',     column('Type of resource', trim: true)
+to_field 'cho_is_part_of',   column('Yale Collection', trim: true)
+to_field 'wr_dc_rights',     column('Rights', trim: true)
```
anarchivist: If this is about the digital image or the web resource itself, it should be it should be associated with `agg_is_shown_at`, `agg_is_shown_by`, `agg_preview`, etc.

lib/traject/yale_csv_config.rb
```diff
@@ -0,0 +1,46 @@
...
+
+settings do
+  provide 'writer_class_name', 'DlmeJsonResourceWriter'
+  provide 'reader_class_name', 'CsvReader'
+end
+
+to_field 'agg_provider',     provider
+to_field 'id',               normalize_prefixed_id('OrbisBibID')
+to_field 'cho_identifier',   column('OrbisBibID', trim: true)
+to_field 'cho_provenance',   column('Place of origin', trim: true)
```
anarchivist: This seems like place of publication information. Is this the right mapping?

lib/traject/yale_csv_config.rb
cmh2166: it should be. was commented on in the spreadsheet mapping as well. i think wayne is trying to differentiate between key dates + provided / raw / text dates, but we don't have support for that in the IR MAP right now.

lib/traject/yale_csv_config.rb
cmh2166: The values are tricky here - there is analysis on the values in the spreadsheet that support the odd mapping (if i recall correctly).

PR #290, opened 2017-09-05
Update docs

cmh2166:
- [x] #76 Complete & include dataflow diagram. 
- [ ] #76 Include Traject logic / override steps / workflow
- [x] #74 Won't solve, but add comment about in docs (data decision area)
- [x] #88 Add Steps for including records & mappings for local testing before inclusion
- [x] #98 Won't solve, but add comment about in docs (data decision area)
- [x] Include logic in data decisions area for current URL look-ups
- [x] #245 review + add to #285 (generic configs)
- [ ] #246 create docs on macros, when to add to, logic therein

Will not get through all of this, and there are other docs questions, but lining up for work.

PR #291, opened 2017-09-05
Do a nil check on the node prior to tranforming it.

### LINKED ISSUE #289: TEI attribute look up needs handling for nil
TEI attributes for language needs to have handling for empty response ([]).

jcoyne:
Ref https://app.honeybadger.io/projects/53082/faults/34547776

Fixes #289


cbeer:
It looks like @cmh2166 also fixed this in #290. I think you two should discuss and pick an approach.

jcoyne:
@cmh2166 I prefer this approach as it only runs the query once.  What do you think?

PR #292, opened 2017-09-06
Limit Princeton Movie Poster indexing to MODS

waynegraham:
We received METS data that includes paths for the thumbs and full-size images. This limits the indexing to just the MODS records for now.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -58,6 +58,7 @@ import:
         inst_id: met
     princeton_mods:
       directory: movie-posters/records/princeton
+      only: !ruby/regexp '/(.*)+.mods/'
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -58,6 +58,7 @@ import:
         inst_id: met
     princeton_mods:
       directory: movie-posters/records/princeton
+      only: !ruby/regexp '/(.*)+.mods/'
```
jcoyne: I think you want `/\.mods$/` (strings that end with `.mods`.  If you wanted to match the whole string `/*.\.mods$/` would be the way to go.

PR #363, opened 2019-02-06
updating solr wrapper version

jacobthill:
updating readme doc

add bnf traject config

add sources for bnf data

config bnf-ideo

normalize contributor

strip white space

language translation map

update mappings

update name tranformation

change cho_has_type to cho_type

PR #365, opened 2019-02-06
Remove .DS_Store files

jcoyne:
Reverts part of #363


peetucket:
do we need to add something to .gitignore to keep them from coming back by mistake?

jcoyne:
@peetucket no, this should be in your global gitignore: https://help.github.com/articles/ignoring-files/

peetucket:
Ok, worked with Jacob to add into global gitignore

PR #368, opened 2019-02-06
Rubocop the traject configs

jcoyne:
Reverts part of #363


peetucket:
so this is the rubocop autogen instead of blindly excluding all traject stuff?  which means all future traject stuff will need to comply?

jcoyne:
@peetucket not entirely.  I moved the `NS` declaration to a mixin and wrapped some long lines. We have to just ignore the mixin usage.

PR #370, opened 2019-02-07
readded namespace to bnf config files

jacobthill:
Removing the NS caused this error: 
```
Error loading configuration file lib/traject/bnf_ideo_config.rb:20 NameError:uninitialized constant NS
lib/traject/bnf_ideo_config.rb:20:in `block in load_config_file'
```
If I remove the NS parameter from extract_xml I get this error:
```
Error loading configuration file lib/traject/bnf_ideo_config.rb:20 ArgumentError:wrong number of arguments (given 1, expected 2..3)
/Users/jtim/Dropbox/DLSS/DLME/dlme/lib/traject/macros/xml.rb:11:in `extract_xml'
lib/traject/bnf_ideo_config.rb:20:in `block in load_config_file'
```
I need help resolving this

### DIFF
--- a/lib/traject/bnf_bo_config.rb
+++ b/lib/traject/bnf_bo_config.rb
@@ -16,6 +16,8 @@ end
 
 record = 'srw:record/srw:recordData/oai_dc:dc'
 
+NS = { srw: "http://www.loc.gov/zing/srw/", oai_dc: "http://www.openarchives.org/OAI/2.0/oai_dc/", dc: "http://purl.org/dc/elements/1.1/"}
+
 # Cho Required
 to_field 'id', extract_xml("#{record}/dc:identifier", NS), strip
 to_field 'cho_title', extract_xml("#{record}/dc:title", NS), strip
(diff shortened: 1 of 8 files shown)


jcoyne:
Can you add a test so that we don't see this problem in the future?

jcoyne:
My intention here was to move NS to Macros::DLME::NS.  I don't think that made it in the configs. Could we do that instead, so we don't duplicate this in 7 places?

peetucket:
> My intention here was to move NS to Macros::DLME::NS. I don't think that made it in the configs. Could we do that instead, so we don't duplicate this in 7 places?

Something like the last commit I did here?  

https://github.com/sul-dlss/dlme/blob/683dd3000b0657441358a4a3a22a732f5eea42e6/lib/traject/macros/dlme.rb#L6-L10

I named it `LOC_NS` to distinguish from other NS declarations in other macros (e.g. https://github.com/sul-dlss/dlme/blob/master/lib/traject/macros/fgdc.rb#L6) that appear to be specific to other extractors.  I'm presuming this namespace may be used for other things than BNF and belongs in a more global macro.

PR #372, opened 2019-02-07
fix search and replace error in translation map

jacobthill:
@peetucket search and replace added LOC_NS in the middle of translation map.


jcoyne:
😆

peetucket:
oops

PR #385, opened 2019-02-12
The application should not be dependent on a database at startup

jcoyne:
This can occur if the app starts up ahead of the database or if we just want to compile assets

See: https://github.com/projectblacklight/spotlight/issues/2133


justinlittman:
Does this create a problem if app started before db ready?

PR #387, opened 2019-02-12
Removed unused postgres volume from docker-compose.

justinlittman:


jcoyne:
@justinlittman rebase please.

PR #397, opened 2019-02-12
Remove cable container from docker-compose.

justinlittman:


docker-compose.yml
```diff
@@ -65,19 +65,6 @@ services:
       - '.env'
     networks:
       - dlme
-  cable:
-    depends_on:
-      - 'redis'
-    build: .
-    command: puma -p 28080 cable/config.ru
```
jcoyne: Should we remove this file too?

justinlittman: Done @jcoyne.

PR #383, opened 2019-02-12
Update rubocop to 0.63.1

jcoyne:

PR #399, opened 2019-02-12
remove traject and other unnecessary stuff from rubocop_todo.yml

peetucket:


peetucket:
oops, duplicates what is in #383

PR #407, opened 2019-02-13
Added ability to fetch DLME JSON records by URL.

### LINKED ISSUE #342: Add form input on webapp to use an S3 URI to fetch data
I will know this is done when:
* [ ] the app accepts a URI to fetch data
* [ ] the data is fetched and loaded into Spotlight

justinlittman:
closes #342


app/controllers/s3_harvester_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      flash[:error] = t('spotlight.resources.fetch.error')
+      return redirect_to spotlight.new_exhibit_resource_path(current_exhibit)
+    end
+    items = resp.body.split("\n")
+    record_count = 0
+    items.each_with_index do |item, index|
+      begin
+        object = DlmeJson.create!(data: { json: item }, exhibit: current_exhibit)
+        object.reindex_later
+      rescue ActiveRecord::RecordInvalid
```
jcoyne: Rather than using rescue for flow control, it would be better to do:
```ruby
object = DlmeJson.new(data: ...)
if object.valid?
  # handle invalid
else
  object.save_and_index
end
```

justinlittman:
To test in running app:
1. Go to http://localhost:3000/library/catalog/admin
2. Click Add Items.
3. Click DLME S3 Fetch.
4. Enter a URL (e.g., https://gist.githubusercontent.com/justinlittman/2274e921a124a5839406823e6df20ef5/raw/eb80858c0caa4fe32cc8a3414d2c635bec5ffc83/gistfile1.txt#) and click Fetch.

Expected result:
You are notified that records were successfully added.

app/controllers/s3_harvester_controller.rb
justinlittman: Done

app/controllers/s3_harvester_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    items.each_with_index do |item, index|
+      object = DlmeJson.new(data: { json: item }, exhibit: current_exhibit)
+      unless object.valid?
+        flash[:error] = t('spotlight.resources.create.error', index: index + 1)
+        return redirect_to spotlight.new_exhibit_resource_path(current_exhibit)
+      end
+      object.save_and_index
+      record_count = index + 1
+    end
+    flash[:notice] = t('spotlight.resources.create.success', index: record_count)
```
jcoyne: Why not just use items.count?

justinlittman: Hah, that'll work. Fixing.

config/locales/en.yml
```diff
@@ -52,3 +56,12 @@ en:
     search:
       fields:
         cho_title_ssim: Title
+    resources:
+      upload:
+        form:
+          cho_title_ssim: Title
```
jcoyne: Are you sure we need this line?  was this the result of a rebase?

justinlittman: Yes, from rebase. Which line #s do you think can go?

jcoyne: 61 & 62 for certain.

justinlittman: OK. Removed.

PR #412, opened 2019-02-14
Remove the vestiges of the harvester and traject

jcoyne:


jcoyne:
@peetucket That's a question for @anarchivist and @jacobthill.  Do we want to keep our old harvest data?

anarchivist:
We can get rid of the harvest data as persisted in the database (i.e., that associated with the harvest models and the migrations @peetucket references). However, we need to ensure that we can re-run the transformations for the data we'd previously aggregated.

jcoyne:
@peetucket are you okay if we move the data destruction to a different PR?

PR #417, opened 2019-02-19
edit application-profile

jacobthill:
This seems clearer to me. Let me know if you think it should change or stay as it was before.


anarchivist:
Awesome. Thank you!

PR #418, opened 2019-02-19
change name of translation map

jacobthill:
update translation map name to match file name change in https://github.com/sul-dlss/dlme-transform/commit/7f56cdba84e49068ba6ae29609bb444f9c35a91e


jcoyne:
This file is no longer in the dlme code.

PR #419, opened 2019-02-22
Adds endpoint for receiving transform notifications. Adds model for t…

### LINKED ISSUE #410: Notify the user when a transform is complete.
This message should include the URI for the file in the s3 bucket.  The message should be persistent, so that I can find it again if I dismiss it.

justinlittman:
…ransform notifications. Adds list page for transform notifications.

closes #410

### DIFF
--- a/app/controllers/transform_results_controller.rb
+++ b/app/controllers/transform_results_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+require 'json'
+require 'date'
+
+##
+# Handle notifications of transform
+class TransformResultsController < Spotlight::ApplicationController
+  protect_from_forgery except: :create
+  before_action :authenticate_user!, except: :create
+
+  # This is invoked by SNS HTTP subscription
+  def create
+    TransformResult.create(build_notification)
+    head :ok
+  end
+
+  def show
+    authorize! :show, :transform
+    @results = TransformResult.order(timestamp: :desc).limit(20)
+  end
+
+  private
+
+  def build_notification
+    notification_msg = JSON.parse(JSON.parse(request.raw_post)['Message'])
+    {
+      url: notification_msg['url'],
+      data_path: notification_msg['data_path'],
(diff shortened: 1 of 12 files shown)


app/controllers/transform_results_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+##
+# Handle notifications of transform
+class TransformResultsController < Spotlight::ApplicationController
+  protect_from_forgery except: :create
+  before_action :authenticate_user!, except: :create
+
+  # This is invoked by SNS HTTP subscription
+  # rubocop:disable Metrics/AbcSize
+  def create
+    notification_msg = JSON.parse(JSON.parse(request.raw_post)['Message'])
```
jcoyne: Can this be extracted into`build_notification`?  If that method also did the `DateTime` cast, could we just have `TransformResult.create(build_notification)`?

docker-compose.yml
```diff
@@ -70,6 +77,20 @@ services:
...
+    image: localstack/localstack
+    ports:
+      - "4575:4575"
+      - "4572:4572"
+      - "8080:8080"
+    environment:
+      - SERVICES=sns,s3
+      - DOCKER_HOST=unix:///var/run/docker.sock
+      - DEBUG=1
+#    volumes:
```
jcoyne: remove?

spec/controllers/transform_results_controller_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    before do
+      allow(TransformResult).to receive(:new).and_return(mock_transform_result)
+      allow(mock_transform_result).to receive(:save)
+    end
+
+    # rubocop:disable RSpec/ExampleLength
+    it 'returns 200' do
+      post :create, body: msg
+      expect(TransformResult).to have_received(:new).with(
```
jcoyne: You are checking for `new` and `save` here, but in the actual controller, you're calling `create`.  Any reason?

PR #428, opened 2019-02-25
Set path and ID appropriately for redis

aaron-collier:
Remove auth (password) from redis config as it is not set in AWS.

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,9 +1,9 @@
 if defined? Sidekiq
   Sidekiq.configure_server do |config|
-    config.redis = { url: "redis://:#{ENV['REDIS_PASS']}@#{ENV['REDIS_HOST']}:#{ENV['REDIS_PORT']}/1" }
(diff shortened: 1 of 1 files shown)


config/initializers/sidekiq.rb
```diff
@@ -1,9 +1,9 @@
 if defined? Sidekiq
   Sidekiq.configure_server do |config|
-    config.redis = { url: "redis://:#{ENV['REDIS_PASS']}@#{ENV['REDIS_HOST']}:#{ENV['REDIS_PORT']}/1" }
+    config.redis = { url: "redis://#{ENV['REDIS_HOST']}:#{ENV['REDIS_PORT']}/1", id: "0001" }
```
jcoyne: I'm not familliar with this setting. I've only ever seen it like this (no id): https://github.com/mperham/sidekiq/wiki/Using-Redis#using-an-initializer  Can you explain why we need this?

aaron-collier: The jobs aren't getting picked up without it from what I can tell. This ID comes from the terraform output.

aaron-collier: I'll drop the ID out of this change, as removing the auth is most important...

PR #435, opened 2019-02-27
Allows specification of data_dir to request partial transform.

justinlittman:
closes #443

### DIFF
--- a/app/services/transform_notification.rb
+++ b/app/services/transform_notification.rb
@@ -2,24 +2,18 @@
 
 # Sends SNS messge that says a transform was requested
 class TransformNotification
-  def self.publish
-    new.publish
+  def self.publish(data_dir)
+    new.publish(data_dir)
   end
 
-  def publish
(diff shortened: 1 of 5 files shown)


app/services/transform_notification.rb
```diff
@@ -2,25 +2,19 @@
...
+  def self.publish(data_dir)
+    new.publish(data_dir)
   end
 
-  def publish
+  def publish(data_dir)
     client.publish(
       topic_arn: Settings.sns.topic_arn,
-      message: message
+      message: data_dir
```
jcoyne: I wonder if we should have this be json, so we can send other fields in the future. I'm thinking that a request identifier would be good, so we can track things we kick off, but that never return.  Is that premature?

justinlittman: Yeah, I thought about that but that would require that the lambda parse the message as JSON which is just an extra iota of complexity. Could always add later.

PR #438, opened 2019-02-28
update cho_edm_type to reflect actual use

jacobthill:
@anarchivist I added the types that we are currently using or that I anticipate using to the application profile and took the values out that of the translation map that we are not/will not likely use https://github.com/sul-dlss/dlme-transform/pull/84. When these two pull requests are merged, the cho_edm_type should be used the way it is documented in the application profile. I will ignore the other type fields for now since we haven't figured out exactly how to proceed yet. Let me know if this makes sense or if you have any suggestions.


jacobthill:
I think that does it. I noticed that "collection" was lingering in the types.yaml so I removed that. Let me know if there's anything else to do on this.

anarchivist:
I think you still need to pull out `cartographic`. Also, can you rebase against master?

jacobthill:
@anarchivist I think I got it this time

Issue #437, opened 2019-02-28
Update docs


justinlittman:
https://github.com/sul-dlss/dlme/tree/master/docs should either be updated or removed or a bit of both.

anarchivist:
Keep:
* Application profile and model diagram
* Framing statement for prototype

Replace/Edit:
* docs/Readme content (refers to prototype)
* Data flow diagram
* top-level readme

Move:
* Add data source (to dlme-harvest?) 
* basic configs (to dlme-traject?)

PR #441, opened 2019-02-28
Update dataflow diagram

jcoyne:
Ref #437


jcoyne:
This got combined into #442

PR #450, opened 2019-03-01
Refactor the TransformController

jcoyne:
Rails knows how to parse JSON responses.  Also changed to a request
test.


jcoyne:
@justinlittman is it possible it's not setting the `Content-type` header?

jcoyne:
https://forums.aws.amazon.com/thread.jspa?threadID=69413 😢

justinlittman:
Possible, but not according to the docs: https://docs.aws.amazon.com/sns/latest/dg/sns-http-https-endpoint-as-subscriber.html

justinlittman:
Hah, a thread about a minor bug that started in 2011.

PR #459, opened 2019-03-04
Fixes to sidekip Dockerfile.

justinlittman:


docker/Dockerfile.worker
```diff
@@ -11,7 +11,7 @@ ARG SECRET_KEY_BASE
 # Create and set the working directory as /opt
 WORKDIR /opt
 
-RUN apk add --no-cache build-base libxml2-dev libxslt-dev postgresql-dev \
+RUN apk add --no-cache build-base libxml2-dev libxslt-dev postgresql-dev gmp-dev sqlite-dev \
```
jcoyne: Rather than add sqlite-dev, we can do #457

PR #460, opened 2019-03-04
Add a dev dockerfile

jcoyne:
The main dockerfile should be for `RAILS_ENV=production` and thus should be bundled `--without development test`

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -4,7 +4,7 @@ FROM starefossen/ruby-node:2-10
 LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
 
 # Set default RAILS environment
-ENV RAILS_ENV=production
+ENV RAILS_ENV=development
 ENV RAILS_SERVE_STATIC_FILES=true
 ENV BUNDLER_VERSION=2.0.1
 ARG SECRET_KEY_BASE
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.dev
```diff
@@ -0,0 +1,29 @@
+# This image's actual base image
+FROM starefossen/ruby-node:2-10
+
+LABEL maintainer="Aaron Collier <aaron.collier@stanford.edu>"
+
+# Set default RAILS environment
+ENV RAILS_ENV=production
```
justinlittman: Why is this production?

justinlittman:
Thinking about this more. Would need to do the same for worker as well. Is is worth maintaining multiple sets of duplicate Dockerfiles just to avoid dev and test deps?

jcoyne:
@justinlittman I've had problems before running `RAILS_ENV=production` when you've got dev and test deps installed. See https://blog.kurttomlinson.com/posts/don-t-install-the-spring-gem-on-your-production-rails-server

justinlittman:
Can we handle the difference with a build arg? Trying to avoid repeating / confusing ourselves.

jcoyne:
@justinlittman I don't know of any way, but maybe you do?

justinlittman:
No, but let me give it a crack.

PR #471, opened 2019-03-06
Switches transform message to json to support sending blank data_dir.

### LINKED ISSUE #468: Dev: trying to run all transforms leads to error:
Steps to reproduce

* Go into dev -> Site administration -> Transform data
* Leave path field blank
* Click "Begin transform"
* Receive "We're sorry, but something went wrong."

Seems to trigger the following: https://app.honeybadger.io/projects/53082/faults/46255424 (see #432)

justinlittman:
closes #468

### DIFF
--- a/app/services/transform_notification.rb
+++ b/app/services/transform_notification.rb
@@ -9,7 +9,7 @@ class TransformNotification
   def publish(data_dir)
     client.publish(
       topic_arn: Settings.sns.topic_arn,
-      message: data_dir
+      message: message(data_dir)
     )
   end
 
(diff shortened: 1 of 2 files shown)


app/services/transform_notification.rb
```diff
@@ -18,4 +18,8 @@ def publish(data_dir)
   def client
     Aws::SNS::Client.new
   end
+
+  def message(data_dir)
+    JSON.generate(data_dir: data_dir || nil)
```
jcoyne: I'm not sure I understand what the `data_dir || nil` would accomplish.  is this casting `false` to `nil`?  Do we need to consider the case of an empty string?

justinlittman: It does nothing in Ruby. But it works great in Python.

The lambda will handle it being either null or blank.

PR #476, opened 2019-03-08
Switches to background loading of resources from S3.

### LINKED ISSUE #472: Parse the IR in the background
When a large IR file is selected, it may take more than 30s to parse and load it into the database.  Putting this in a background job will ensure the web user gets a response in a reasonable time.

justinlittman:
closes #472

### DIFF
--- a/app/jobs/fetch_resources_job.rb
+++ b/app/jobs/fetch_resources_job.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# Load JSON resources from a url, persist them, and index them
+class FetchResourcesJob < ApplicationJob
+  queue_as :default
+
+  def perform(url, exhibit)
+    resp = Faraday.get(url)
+
+    resource = resp.body.split("\n")
+    resource.each_with_index do |item, index|
+      create_or_update_resource(item, exhibit, index)
+    end
+    logger.info("#{resource.count} records were created from #{url}.")
+  end
+
+  private
+
+  def create_or_update_resource(item, exhibit, index)
+    json = JSON.parse(item)
+    resource = DlmeJson.find_or_initialize_by(url: json['id'], exhibit: exhibit)
+    resource.data = { json: item }
+
+    raise "Resource #{index + 1} in #{url} is not valid." unless resource.valid?
+
+    resource.save
+    resource.reindex
+  end
+end
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,8 @@ services:
       - RDS_PORT=5432
       - SOLR_URL=http://solr:8983/solr/dlme
       - SECRET_KEY_BASE
+      - REDIS_HOST
+      - REDIS_PORT
       # For development
       - RAILS_SERVE_STATIC_FILES=true
       - RAILS_ENV=development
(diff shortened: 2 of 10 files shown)


docker-compose.yml
```diff
@@ -27,8 +29,8 @@ services:
     networks:
       - dlme
     # For running in dev
-    volumes:
-      - ".:/opt/"
+#    volumes:
+#      - ".:/opt/"
```
jcoyne: If we don't have this, then if we change a file on the fs, isn't not changed on the server, right?

app/jobs/fetch_resources_job.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  private
+
+  def create_or_update_resource(item, exhibit, index)
+    json = JSON.parse(item)
+    resource = DlmeJson.find_or_initialize_by(url: json['id'], exhibit: exhibit)
+    resource.data = { json: item }
+
+    raise "Resource #{index + 1} in #{url} is not valid." unless resource.valid?
```
jcoyne: The user of the website is never going to see this error.   Do we need to write some persistent notification the database (could be a new ticket)?

docker-compose.yml
justinlittman: Yes. So if someone wants to run the app in a dev mode, the appropriate lines could be uncommented. But open to other suggestions.

app/jobs/fetch_resources_job.rb
justinlittman: I figured it would be Honeybadger. Since all data should already be validated IR, I would expect failures to be bugs.

app/jobs/fetch_resources_job.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    resource.save_and_index
```
jcoyne: `save_and_index` is an async operation. Since this is a background job, maybe it should be synchronous here? See https://github.com/projectblacklight/spotlight/blob/3b30e827ba40772d519ec5123f5b8c7dd945131f/app/models/spotlight/resource.rb#L27-L29

docker-compose.yml
```diff
@@ -16,9 +16,11 @@ services:
       - RDS_PORT=5432
       - SOLR_URL=http://solr:8983/solr/dlme
       - SECRET_KEY_BASE
+      - REDIS_HOST
+      - REDIS_PORT
       # For development
-      - RAILS_SERVE_STATIC_FILES=true
-      - RAILS_ENV=development
+#      - RAILS_SERVE_STATIC_FILES=true
+#      - RAILS_ENV=development
```
jcoyne: If we don't set development here, then it's running in production mode, right?  Why?

justinlittman: Because otherwise the sidekiq config doesn't work. Would take some additional tweaking.

docker-compose.yml
jcoyne: I would prefer that a change like this should be made in a separate pull request that is about changing the environment.  I think the change you want to make here is adding these lines to `config/environments/development.rb`


```ruby
config.active_job.queue_adapter     = :sidekiq
config.active_job.queue_name_prefix = "dlme_#{Rails.env}"
```

docker-compose.yml
jcoyne: Because `config/environments/development.rb` doesn't set a queue_adapter.

docker-compose.yml
justinlittman: Agree.

PR #487, opened 2019-03-11
Add build arg to control whether dev and test deps are bundle installed.

justinlittman:


docker/Dockerfile
```diff
@@ -20,7 +21,8 @@ RUN gem install bundler
 COPY Gemfile /opt
 COPY Gemfile.lock /opt
 
-RUN bundle install
+RUN bundle config build.nokogiri --use-system-libraries \
+    && bundle install --without ${BUNDLE_WITHOUT}
 
 COPY . .
 RUN rake assets:precompile
```
jcoyne: This step is unnecessary in development mode.

justinlittman: It found the uglifier problem. So while it does add some additional image build time, I'd suggest that it is worth it to have dev as similar to prod as possible.

jcoyne: But it only will find it when you docker build. (and you don't need to do docker build unless you change the Gemfile.lock). Our CI does docker-build and should find such problems.

justinlittman: Still not clear on all this Rails, but doesn't this mean when Gemfile.lock changes and you need to do a rebuild, it will take an extra 30 seconds to run `rake assets:precompile`? Or am I misunderstanding?

jcoyne: yes, it will take an extra 30 s to run `rake assets:precompile`, but since it's in development mode, it won't use the output of that. It's going to recompile the assets on every request.

aaron-collier:
In general, I'm in favor of this approach over multiple files, and :+1: to it... have one small comment that may be unnecessary.

PR #510, opened 2019-03-29
Create a way to delete objects

### LINKED ISSUE #489: Ensure we have a way to deal with deleted records
Occasionally, a reharvest will mean that identifiers change, or that we are no longer pulling in records from a given data source. We need to identify a way to handle deletions that is exposed in the webapp.

* [x] Undertake analysis of what's in Spotlight now
* [x] Discuss requirements with @jacobthill
* [ ] Implement it

## Use case / Story

As the DLME manager I need to be able to delete metadata for an entire institution or collection that has been imported.

## Request

By pointing at the same folder used for transformations, use the metadata in the directory structure to find existing resources and remove them.

## Follow up questions / concerns

- [ ] Verify that changes in metadata do not cause duplication.
- [ ] Verify what happens when the same data set has been added more than once
- [ ] Identify duplicate records for removal
- [ ] The currently available public/private flag may be a suitable option (assuming it can be applied in bulk as described above) but the ideal solution is record removal.
- [ ] While providing a delete button/option at the record level may be nice, the known use case is bulk. Individual delete will be ticketed separately.

jcoyne:
Fixes #489

### DIFF
--- a/app/jobs/delete_resources_job.rb
+++ b/app/jobs/delete_resources_job.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# Load JSON resources from a url, persist them, and index them
+class DeleteResourcesJob < ApplicationJob
+  queue_as :default
+
+  def perform(url, exhibit)
+    resp = Faraday.get(url)
+
+    resource = resp.body.split("\n")
+    resource.each_with_index do |item, index|
+      remove_resource(item, exhibit, index)
+    end
+    logger.info("#{resource.count} records were removed from #{url}.")
+  end
+
+  private
+
+  def remove_resource(item, exhibit, _index)
+    json = JSON.parse(item)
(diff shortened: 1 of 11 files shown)


app/jobs/delete_resources_job.rb
```diff
@@ -0,0 +1,39 @@
...
+  queue_as :default
+
+  def perform(url, exhibit)
+    resp = Faraday.get(url)
+
+    resource = resp.body.split("\n")
+    resource.each_with_index do |item, index|
+      remove_resource(item, exhibit, index)
+    end
+    logger.info("#{resource.count} records were created from #{url}.")
```
aaron-collier: Should this be "...records were removed from..."?

jcoyne: Yes! Thank you Aaron. That's fixed.

PR #517, opened 2019-04-04
Add download of p12 key from S3 in spotlight_initializer

aaron-collier:
Since we are running this in a container, we want to avoid including the PKCS12 key for google analytics in the public image. This allows us to pass a path to create the file and an S3 path to download the file in ENV variables and then grab the key file during initialization.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -2,8 +2,13 @@ application:
   default_host: ~
   default_from: ~
 
+# pkcs12_key: URI to key file location in S3
+# pkcs12_key_path: Path that will be used to store the PKCS12 Key file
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -4,6 +4,9 @@ application:
 
 analytics:
   web_property_id: ~
+  pkcs12_key: ~
```
jcoyne: Can you document the expected type here?  I'm guessing this is a URI to a key that should be in a private s3 bucket?

aaron-collier: Added a comment above the block.

PR #610, opened 2019-10-30
Only require sidekiq + redis if there's a REDIS_HOST set

cbeer:


config/environments/development.rb
```diff
@@ -28,8 +28,10 @@
   end
 
   # Use a real queuing backend for Active Job (and separate queues per environment)
-  config.active_job.queue_adapter     = :sidekiq
-  config.active_job.queue_name_prefix = "dlme_#{Rails.env}"
+  if ENV['REDIS_HOST']
```
jkeck: Thank you!

PR #625, opened 2019-10-31
Update to Blacklight 7 / Bootstrap 4

cbeer:
We're still tracking some upstream aesthetic issues (icons, some contrast issues) in https://github.com/projectblacklight/spotlight/issues/2224, but functionally this seems to work just fine.


camillevilla:
There are still some Bootstrap 4 fixes being worked on in upstream Spotlight (projectblacklight/spotlight#2224), but this looks like a good start.

Issue #612, opened 2019-10-30
Consider local dev experience improvements


jkeck:
Look into ways to make local development easier + possible without running docker (or redis / postgres / etc)

- [x] https://github.com/sul-dlss/dlme/pull/610

jkeck:
I'm going to take a look at if postgres is required.  It looks like it's not but I want to say that @ggeisler needed to install it in order to get the app up and running.  I'll validate that.

jkeck:
FWIW, at the current state of the `master` branch, you not longer need to have redis/sidekiq running (although can be nice for indexing/debugging) and postgres is not required (it seems to run fine on sqlite).

ggeisler:
@jkeck In case it helps, I believe my issue was:

1. Ran bundle install
2. Got an error when it was trying to install `pg 1.1.4`
3. Installed postgres with Homebrew
4. Re-ran bundle install, this time successfully

So it seems like step 3 was necessary, but it would be good for someone else to verify that, for sure.

cbeer:
I /think/ you should be able to run `bundle install --without production` and not need postgres.

PR #631, opened 2019-11-01
Update README with details about how to run outside of Docker.

jkeck:
Cloes #612

## Why was this change made?
To update the README w/ information about recent changes to the local development environment.

Yes, in fact, that's all this change is! 📚

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,41 @@ And these database configuration settings:
 
 ## Local Development
 
-Requires docker.
+### Without Docker
+The DLME application can be run in a standard ruby/rails environment locally using SolrWrapper, sqlite3, etc.
 
+In order to not have to use Postgres locally (and use sqlite for for development), you'll want to bundle install without the production dependencies.
+
+```bash
+$ bundle install --without production
+```
+
+You can run the standard rails setup script which will run your database migrations, etc.
+
+```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,8 +37,41 @@ And these database configuration settings:
 
 ## Local Development
 
-Requires docker.
+### Local Application Development
```
mjgiarlo: Not going to block changes since I'm not on this work cycle, but I wonder if you would consider changing a couple labels in this change. (I'm not clear on how docker-based work is not local, I guess.) In specific, would you consider changing the `Local Application Development` subheader to `Local Development (Without Docker)`, and `Local E2E Development` to `Local Development (Using Docker)`, or some such?

jkeck: The docker section is a heading under local development (so there are sections under local development, one for w/ Docker, and one w/o).  This was not to indicate that working w/ Docker was not local, but that there was two modalities.

That being said, happy to make this change.

mjgiarlo: > yo dawg, i herd u like local development

Thx, @jkeck!

mjgiarlo:
I'm not approving this PR, cuz not my work cycle, but I want do approve **of** the change just made. Much obliged. :beers:

PR #641, opened 2019-11-06
Also index multilingual fields into a common field

cbeer:
## Why was this change made?

In order to keep the app functional while fields are being transformed into language-tagged fields, keep on indexing them into the same old fields. This also is useful for some analytics and curation queries (e.g. show me all the things without titles)

### DIFF
--- a/app/resource_builders/dlme_json_resource_builder.rb
+++ b/app/resource_builders/dlme_json_resource_builder.rb
@@ -28,6 +28,8 @@ class DlmeJsonResourceBuilder < Spotlight::SolrDocumentBuilder
         # Handle hashes specially because if a hash slips through to Solr, Solr
         # gets unhappy and will return a 400 error.
         if source[key].is_a?(Hash)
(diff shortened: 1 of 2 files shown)


app/resource_builders/dlme_json_resource_builder.rb
```diff
@@ -28,6 +28,8 @@ def to_solr
         # Handle hashes specially because if a hash slips through to Solr, Solr
         # gets unhappy and will return a 400 error.
         if source[key].is_a?(Hash)
+          sink["#{key}_tsim"] = source[key].values.flatten
```
jkeck: One quick question about this (wrt backwards compatibility) since this is a `_tsim` field we could use this in places like the Empty fields facet query, but would this end up being problematic if we wanted to use the field as a facet? (or maybe there's copyfield stuff going on under the hood that I don't grok).

cbeer: The ssim fields happen later (line 57). I'm not totally clear on the separation here, but 🤷‍♂

Issue #630, opened 2019-11-01
Update facet and index fields for multi-language index


aaron-collier:
Example SOLR document:

```
       "id":"themaghribpodcast.podbean.com/l%e2%80%99architecture-de-la-contre-revolution-l%e2%80%99armee-francaise-dans-le-nord-de-l%e2%80%99algerie-c292151bc2d9ddc75279aec57bdd380a",
        "exhibit_library_public_bsi":true,
        "spotlight_exhibit_slug_library_bsi":true,
        "spotlight_exhibit_slugs_ssim":["library"],
        "spotlight_resource_id_ssim":["gid://dlme/DlmeJson/189139"],
        "spotlight_resource_type_ssim":["dlme_jsons"],
[...]

Issue #589, opened 2019-10-02
Create view for mixed language metadata


jacobthill:
Some records will have metadata that is all English ([example](https://spotlight.stage.dlmenetwork.org/library/catalog/ans_2010-44-2)). Some records will have modern Turkish (e.g. Turkish in Latin script) or French, etc. and English ([example](https://spotlight.stage.dlmenetwork.org/library/catalog/oai:cdm21044.contentdm.oclc.org:emirgan%2F925)). Some records will have Arabic and English ([example one](https://spotlight.stage.dlmenetwork.org/library/catalog/010740758), [example two](https://spotlight.stage.dlmenetwork.org/library/catalog/b17704236), [example three](https://spotlight.stage.dlmenetwork.org/library/catalog/oai:cdm15795.contentdm.oclc.org:p15795coll19%2F28)). Eventually, some records will have all Arabic metadata but we don't have an example until QNL updates their metadata. We need a view that will handle these use cases and the problems they pose, including right-to-left and left-to-right languages in the same record. 

The cases with Arabic and English mixed will be among the most challenging. Some of the fields will have Arabic and English for every record (agg_provider => agg_provider_ar, agg_data_provider => agg_data_provider_ar, cho_edm_type => cho_edm_type_ar, cho_has_type => cho_has_type_ar, cho_language => cho_language_ar, cho_date_range_norm => cho_date_range_norm_ar). Some fields could come in either or both languages (titles, contributors, creators, and descriptions are good examples).  When in the Arabic view, Arabic values should be displayed and English equivalents should be suppressed. English values with no Arabic equivalent should still be displayed alongside the Arabic however. The opposite is true as well. When in the English view, English values should be displayed and Arabic equivalents should be suppressed. Arabic values with no English equivalent should still be displayed alongside the English. This is one possible solution though it might be complicated to implement. Another option would be to display all values in the metadata but put the Arabic to the right and the English on the same line to the left (e.g. two columns). This would be easier to implement but we need to determine if it is visually appealing.

ggeisler:
Below is a proposed strategy for how to determine which language-versions of metadata values to display on item show pages when the user is in English and Arabic locales. @jacobthill and I discussed this but I revised the mockups after our discussion so @jacobthill please review and let me know if you see any problems.

Some general notes:
- I think we will want to use the same strategy for result item metadata but that might be a separate issue to look at later.
[...]

PR #642, opened 2019-11-06
Add initial (basic) support for multilingual fields

jkeck:
Part of #630 

## Why was this change made?
So we can begin to see the multilingual content that was indexed transformed/indexed before #589 is ready to be worked on (which will be the more complex logic of what is displayed when certain locales are chosen).

n/a

Note: This is more-or-less just a stop gap so that we can render the multilingual data that is available.  There is work going on both in DLME (for Empty values faceting purposes) and upstream in Blacklight (for features that will support multilingual field rendering) that will likely supersede this code in the long run

## Before
facets-before

result-record-before

## After 
facets-after

result-record-after

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -31,15 +31,28 @@ class CatalogController < ApplicationController
     config.document_solr_path = 'get'
     config.document_unique_id_param = 'ids'
 
-    # solr field configuration for search results/index views
-    config.index.title_field = 'cho_title_ssim'
+    locale_encoded_fields = lambda do |field_prefix, suffix = 'ssim'|
+      (Settings.acceptable_bcp47_codes << 'none').map do |code|
+        [code, "#{field_prefix}.#{code}_#{suffix}"]
+      end.to_h
+    end
+
+    config.index.title_field = locale_encoded_fields.call('cho_title').values
     config.index.thumbnail_field = 'agg_preview.wr_id_ssim'
     config.index.default_thumbnail = 'default.png'
 
-    config.add_index_field 'title', field: 'cho_title_ssim'
+    locale_encoded_fields.call('cho_title').each do |code, field|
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -66,10 +85,14 @@ class CatalogController < ApplicationController
     config.add_facet_field 'contributor', field: 'cho_contributor_ssim', limit: true
     config.add_facet_field 'medium',      field: 'cho_medium_ssim', limit: true
     config.add_facet_field 'dc_rights',   field: 'cho_dc_rights_ssim', limit: true
-    config.add_facet_field 'holding_institution', field: 'agg_data_provider_ssim', limit: true
+    locale_encoded_fields.call('agg_data_provider').each do |code, field|
```
jkeck: @jacobthill is this okay in the interim for the facets? (You can see the before/after screenshots of the facets attached to this PR for a visual example).

app/controllers/catalog_controller.rb
```diff
@@ -31,14 +31,33 @@ class CatalogController < ApplicationController
...
 
-    # solr field configuration for search results/index views
-    config.index.title_field = 'cho_title_ssim'
+    locale_encoded_fields = lambda do |field_prefix, suffix = 'ssim'|
+      (Settings.acceptable_bcp47_codes << 'none').map do |code|
+        [code, "#{field_prefix}.#{code}_#{suffix}"]
+      end.to_h
+    end
+
+    negative_range_query_for_language_fields = lambda do |field_prefix, suffix = 'ssim'|
```
jkeck: I kept this in a separate commit so it can easily be popped off when #641 is merged (but I had already written the code when @cbeer did that, so why not)

jkeck: Removed this commit so we don't have this query going on any longer (and the Empty values facet will be correct after content is reindexed).

PR #669, opened 2019-11-09
Add a Statistics MainNavigation item and page for exhibits.

jkeck:
## Why was this change made?
Because this

n/a

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,12 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+# Add a Statistics MainNavigation item for every exhibit that does not have one
+Spotlight::Exhibit.find_each do |exhibit|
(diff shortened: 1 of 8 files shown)


db/seeds.rb
```diff
@@ -5,3 +5,12 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+# Add a Statistics MainNavigation item for every exhibit that does not have one
+Spotlight::Exhibit.find_each do |exhibit|
```
jkeck: One thing about doing this in the seed file, is I'm not sure that this is run when deploying to AWS.  Maybe @cbeer / @aaron-collier maybe would know more.  This is done in an idempotent way so we could add it on deploy but not sure if we should go that route or do something else.

cbeer: I think you're right, it isn't run automatically. A migration would fix up the current exhibits, but as a one-off  it'd probably be just as easy to add it from the console. 🤷‍♂

jkeck: To close the loop here, I'll work w/ @aaron-collier to console in an run run the migration (or however we would want to do this in AWS land)

Issue #531, opened 2019-07-03
Public-facing view for exhibit analytics/statistics


jacobthill:
Need a public facing data dashboard to show statistics about the current state of DLME. Is seems that the data needed can be pulled from the intermediate representation (IR) via solr query or from Google Analytics (GA).

Requirements:
- [ ] Design work
- [ ] IR: total number of contributing institutions (use unique values from agg_data_providor and agg_providor)
- [ ] IR: total number of records from each institution
- [ ] IR: total number of collections from each institution (need to figure out how to get this; its not in the IR)
- [ ] country of each contributing institution 
- [ ] IR: type of each record 
- [ ] IR: language of each record 
- [ ] GA: total number of users
- [ ] GA: country of users

ggeisler:
@jacobthill A few questions about the requirements:

> IR: total number of collections from each institution (stage and prod)

How is a collection defined? In our currently running instances, there is no facet with the label "Collection" so I'm not sure what this is. Is it a grouping of records from a given institution that we keep track of in the IR and just aren't exposing on the site as a facet?

> country of each contributing institution (stage, prod, and non-partner) (depends on

What does this depend on (I think your sentence was cut off)?

ggeisler:
Below is a mockup for an initial version of the public-facing statistics page. I'll make another ticket as a [placeholder for some potential enhancements](#638), but it seems like the essential things we'd want to show are represented in this mockup.

This is not intended to be the final visual design of the page; after I get further with the site visual design (#134) I should be able to make the design of this page look a bit better.

The counts shown in the mockup come from various sources; @jacobthill can either post a follow-up comment in this ticket about where to get the counts, or work with the developer who takes this ticket on via other means to figure out how to get them.

I believe we'd like this page to exist on -prod, -staging, and -dev, although the Users highlight box and Users section is only applicable to -prod.

### A - Link to Statistics page
The page involves things (tables, mainly) we can't do wit Sir Trevor widgets so I'm proposing we:

- [ ] Manually add a main menu item called "Statistics" 
- [ ] Create a static page to contain the Statistics page content ("Statistics" header should be an `<h1 class="page-title">`, I think)

### B - Highlight the three main metrics we want to convey
- I used Bootstrap's `.jumbotron` for the items, contributors, and users highlight boxes, but customized them to remove a bunch of stuff (also, set to only display when viewport is `md` or wider):

```html
<div class="row d-none d-md-flex justify-content-around">
[...]

PR #675, opened 2019-11-11
Add the items section to the Statistics dashboard

jkeck:
Part of #531 

## Why was this change made?
To add item details to the statistics page.

Screen Shot 2019-11-14 at 3 52 05 PM
Screen Shot 2019-11-14 at 3 51 30 PM

## TODO
- [x] i18n the strings (holding off to avoid churn/understand doc structure)
- [x] Decide if we want to add a class to represent statistics / migrate `@response` accessors out of view.
- [x] Fix-up pivot facet access/linking (it's pretty sloppy/brittle right now)


app/controllers/catalog_controller.rb
```diff
@@ -107,13 +107,18 @@ class CatalogController < ApplicationController
     config.add_index_field 'spatial', field: 'cho_spatial_ssim'
     config.add_index_field 'temporal', field: 'cho_temporal_ssim'
 
+    Settings.acceptable_bcp47_codes.each do |code|
```
jkeck: I could just explicitly add `en` and `ar-Arab` as below for this instead of looping through all of them. 🤷‍♂️

PR #691, opened 2019-11-15
Add a hierarchical facet (and show field) for the cho_type_facet field

### LINKED ISSUE #552: View for cho_has_type (hierarchical facet)
Depends on https://github.com/sul-dlss/dlme-transform/issues/155. DLME has three type values: `cho_edm_type`, `cho_has_type`, and `cho_type`. The second field is a subcategory of the first field (e.g. one value of `cho_edm_type` is Text which has possible `cho_has_type` values of 'manuscript', 'book', etc.). The view should reflect this hierarchy. `cho_type` is whatever value the raw data provides when we receive it from the data provider. This could be useful as well and should also be useable. These values are not standardized; they could be thought of as tags.

jkeck:
Closes #552 

## Why was this change made?
To add support for a type/sub-type hierarchical facet.

n/a

facet
show-view

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,6 +77,7 @@ gem 'riiif', '~> 2.0'
 gem 'sitemap_generator'
 
 gem 'blacklight-gallery', '>= 0.3.0'
+gem 'blacklight-hierarchy', github: 'sul-dlss/blacklight-hierarchy'
 gem 'blacklight-oembed', '>= 0.1.0'
 gem 'jquery-rails'
 gem 'rails_autolink'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,5 +1,5 @@
 GIT
-  remote: git://github.com/projectblacklight/spotlight.git
+  remote: https://github.com/projectblacklight/spotlight.git
   revision: 2fdace3d7d043eeea6e13a16af34299e1c7a3dad
   specs:
     blacklight-spotlight (3.0.0.alpha)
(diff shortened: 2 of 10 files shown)


Gemfile
```diff
@@ -77,6 +77,7 @@ gem 'riiif', '~> 2.0'
 gem 'sitemap_generator'
 
 gem 'blacklight-gallery', '>= 0.3.0'
+gem 'blacklight-hierarchy', github: 'sul-dlss/blacklight-hierarchy'
```
jkeck: We need to cut a release that is BL 7 compatible.  Happy to do that as part of this.

ggeisler:
@jkect 👍I was just writing this comment when you posted yours.

This is a great start but I was wondering about:

1. The alignment of the item counts. I think this has come up before but I forget what the issue is. Are there complications to getting the counts to be right-aligned so they line up with the non-hierarchical facet counts?

2. Maybe your screenshot is not current but we should be able to collapse/expand the "Sound" label to expose "Interview" and other potential child values, correct?

jkeck:
@ggeisler 
For both 1 and 2 this is what I'm meaning about needing some ❤️. It looks like we've added some styling in SW to do things like float the counts and add the (FontAwesome) icons so we'll likely want to do the same here.

The facet items are togglable, just w/o the icons / other styling it's not obvious from the static screenshot.

jkeck:
Added a little ❤️ 
salt


facet-twiddle

mjgiarlo:
:fire: :fire: :fire:

Issue #676, opened 2019-11-12
Display search results as RTL


ggeisler:
Part of #544. UI text in mockup below is shown as English but would be Arabic when we have integrated all necessary translations.

Although shown in the mockup for context, breadcrumbs are covered in #673 and facets are covered in #674.

### A: Search result labels, constraints, and controls
- [x] Elements that are normally left-aligned in content area are right-aligned ("You searched for" label and prev|next navigation block)
- [x] The colon for the "You searched for" label should be on the left-side of the label
- [x] prev|next navigation block: components are flipped, including the "x - y of z" label
- [x] Constraint controls are flipped so that the delete button is to the left of the label portion
- [x] The "Start over" button is left-aligned
- [x] The grouping of search result view icons and per-page and sort by menus is left-aligned and reversed from normal order
- [x] The icon for the pulldown control on the "sort-by" and "per page" buttons is on the left of the label
- [x] The order of the search result view icons is flipped from normal order
- [x] The "List" icon (the currently selected icon in the mockup below) is flipped horizontally, so the bullets part of the icon is on the right instead of the left. `transform: scaleX(-1);` does this on Chrome, and likely(?) all other browsers we care about. According to [Caniuse](https://caniuse.com/#search=transform) some IE versions don't support `transform` for SVGs, but a transform attribute can be used. Might require a bit of investigation, although it isn't the end of the world if the icon is not always flipped.

The other result view icons do not need to be flipped (they're either more or less symmetrical or a special case (slideshow icon)).

### B: Result items
- [x] Arrangement of result item elements is flipped: thumbnail is right-aligned in main content area, metadata field labels are left-aligned, metadata values are right-aligned
- [x] Result item numbers: The number is right-aligned and the period goes to the left of the number, regardless of whether we stick with Western Arabic numbers or convert to Eastern Arabic numbers, as discussed in #639. (I don't know if we're going to include the default Blacklight item numbering in DLME, but I'm including it in the mockup in case we are doing this work upstream and want to make sure that it's handled for any future adopters that do want to include item numbering.) 
- [x] Result item title: Follows the number and is right-aligned, so if the title wraps it starts the second line at the right-side of the content area

### C: Pagination
- [x] Pagination elements are flipped so that the group is right-aligned and starts (from a RTL perspective) with Previous, etc.

### Mockup - List view

[...]

PR #695, opened 2019-11-18
Fix RTL styling for hierarchical facets

cbeer:
Part of #676 

Before:
Screen Shot 2019-11-18 at 11 41 40

After:
Screen Shot 2019-11-18 at 11 40 08


camillevilla:
Chatted about this with @cbeer and we're deferring kicking RTL things upstream until we can decide on an approach (i.e. do we need a `blacklight-rtl` gem?).

PR #698, opened 2019-11-18
Display both Gregorian and Hijri dates on results and show pages

### LINKED ISSUE #693: Display both Gregorian and Hijri dates on results and show pages
Goal is for the user to see the object's date value according to both the Gregorian and Hijri calendars, when the object data supports this (though I believe all records should have the data to support this). We also want to show, separately, the original or source value for the object date, since it can include values that are meaningful to the user but not translatable between calendar systems.

### Show page examples

Below left are the IR values for the `cho_date` and `cho_date_range_xxxx` fields. The recommendation is to use the `cho_date_range_xxxx` values to populate a Date field, shown first, followed by a "Source date" field that contains the `cho_date` value(s). Below right is an excerpt from the show page metadata for the item using this approach:

Screen Shot 2019-11-18 at 2 16 08 PM

Displaying the `cho_date` value(s) as the "Source date" field is straightforward.

For `cho_date_range_xxxx` we'd need to combine the values from the two source fields and display them on the same line, with the " - ", "(Gregorian)", " / ", and "(Hijri)" text injected.

Here's a slightly different example (value of the `cho_date` field is a bit different) of the source-to-desired-display, but pretty much the same approach (as best I can tell, the source data for these fields across all of our records pretty consistently matches one of these two patterns):

Screen Shot 2019-11-18 at 2 29 02 PM

### Result page example
For a search result item, we want to do pretty much the same thing, only I think we might only want to show the "Date" field based on the `cho_date_range_xxxx` values and omit the "Source date" field. 

But the "Date" value here would be exactly the same as described in the Show page examples above. Here's the result item entry for the first example item above:

Screen Shot 2019-11-18 at 2 15 31 PM

cbeer:
Fixes #693

Screen Shot 2019-11-18 at 14 36 40

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -82,6 +82,13 @@ class CatalogController < ApplicationController
       config.add_index_field 'title', **field_config
     end
 
+    config.add_index_field 'date range', helper_method: :display_date_ranges, values: (lambda do |_field_config, document|
+      {
+        gregorian: document.fetch('cho_date_range_norm_ssim', []).map(&:to_i),
+        hijri: document.fetch('cho_date_range_hijri_ssim', []).map(&:to_i)
+      }
+    end)
+
     config.add_index_field 'date', field: 'cho_date_ssim'
     multilingual_locale_aware_field.call('agg_data_provider') do |field_config|
       config.add_index_field 'holding institution', **field_config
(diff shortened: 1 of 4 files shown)


ggeisler:
Nice. Eventually I think we want to use "Date" and "Source date" as the field labels, but perhaps that is not part of this work (or we just do it via the Config > Metadata label overrides in the UI).

cbeer:
Yeah, I assumed the label work would happen in the curator UI (or done as part of a comprehensive UI relabeling).

PR #706, opened 2019-11-19
Solrize date ranges as integers

### LINKED ISSUE #697: Index date range fields as integers
Both cho_date_range_norm  and cho_date_range_hijri should be indexed as stored integers (`*_isim`) in order to drive the date range plugin.

mjgiarlo:
## Why was this change made?

Solrize date ranges as integers. Fixes #697

Also:
* Use https URLs for GitHub-based repositories
* Update docker-compose configuration 
  1. Remove explicit networks (this is not necessary); and
  2. Use a mounted volume in the database container to store data

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,5 +33,4 @@ config/environments/*.local.yml
 /public/assets/
 
 # Ignore any docker-compose created volume data
-/postgres-data/
 /solr-data/
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 source 'https://rubygems.org'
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -33,5 +33,4 @@ config/environments/*.local.yml
 /public/assets/
 
 # Ignore any docker-compose created volume data
-/postgres-data/
```
ndushay: why is this being removed?  Does this PR need it?

mjgiarlo: It's associated with the addition of the `postgres-data` volume in `docker-compose.yml`. Now this directory won't be created in the `dlme` directory; it will be created in containerland as a mounted volume.

Issue #688, opened 2019-11-15
Add a Contributors section to the Statistics page


jkeck:
Split out from #531

### E - Contributors table, Institution and Country columns
* "Institution" comes from the `agg_provider` field
* "Country" comes from the `agg_provider_country` field

### F - Contributors table, Collections column
* Similarly, some work might need to be done to ensure we have represented and indexed the number of collections associated with each contributing institution.

![Screenshot](https://user-images.githubusercontent.com/101482/68168823-5e280800-ff27-11e9-88f0-20bac4649ab8.png)

PR #708, opened 2019-11-19
Add the jumbotron section for contributors

jkeck:
contribs-jumbotron

In my environment I have the following transforms indexed:
- output-f9967242-f6c2-468e-bb6f-594caed0cf0f.ndjson (from dev)
- output-db22239a-81e4-4118-a667-26fb2b12f889.ndjson (from dev, partially indexed)
- output-12fa7465-adb2-4dc5-b9fd-c302986d45ec.ndjson (from dev, partially indexed)

## Why was this change made?
As a follow on to #688, because, I....uh forgot this part.

### DIFF
--- a/app/views/statistics/show.html.erb
+++ b/app/views/statistics/show.html.erb
@@ -1,13 +1,23 @@
 <h1 class="page-title"><%= t('statistics.header') %></h1>
 
-<div class="row d-none d-md-flex justify-content-around">
-  <div class="jumbotron col-4-md">
+<div class="row d-none d-md-flex justify-content-around text-center">
+  <div class="jumbotron">
     <h2 class="display-6">
       <%= t('statistics.dashboard.header.total_items', count: @statistics_dashboard.items.total) %>
     </h2>
     <hr class="my-4">
-    <p class="text-center">
+    <p>
       <%= t('statistics.dashboard.header.total_types', count: @statistics_dashboard.items.by_type.count) %>
+    </p>
+  </div>
+
+  <div class="jumbotron">
+    <h2 class="display-6">
+      <%= t('statistics.dashboard.header.total_contributors', count: @statistics_dashboard.contributors.total) %>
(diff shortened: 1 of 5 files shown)


app/views/statistics/show.html.erb
```diff
@@ -10,6 +10,16 @@
       <%= t('statistics.dashboard.header.total_types', count: @statistics_dashboard.items.by_type.count) %>
     </p>
   </div>
+
+  <div class="jumbotron col-4-md">
```
cbeer: Is `col-4-md` the right selector here?

ggeisler: Not to answer for @jkeck but I did the mockup of this page as HTML and I believe I used `col-4-md` and it worked fine. There will eventually be three (and only three) boxes. And I proposed to hide the boxes below `md` (they're nice, but not critical). But if you have a reason you think `col-4-md` isn't the best choice I'd be interested to hear it, for my own edification...

jkeck: Yeah, this was copy/pasted from the HTML mockup, but I believe what @cbeer is referring to is that the bootstrap column definition would be `col-md-4` (the column width and the breakpoint are swapped).

So it looks like we're not getting any column definition currently and just getting what `justify-content-around` gives us on the jumbotrons.

The main difference being what width we get on the columns since anything below `md` gets hidden anyway.

## Without a col
no-col

## Using col-md-4
col-md

jkeck: Actually, using col-md-4 w/ 3 jumbtrons doesn't look so good, so I'm pulling the class.

yikes

ggeisler: Ah okay, didn't notice that I transposed those pieces. I just double-checked my HTML mockup and it looks fine with `col-4-md` but that's just because `col-4-md` is meaningless and it's being ignored; my mistake!

PR #709, opened 2019-11-19
Use Mada font (Latin and Arabic scripts) by default for DLME

### LINKED ISSUE #655: Use an Arabic font when the locale is Arabic
Part of #544.

Because Arabic characters are shorter than Latin characters, we should try to use a more Arabic-optimized font when the locale is Arabic. I'm far from an expert on this but from some investigation it seems like the Google font "Amiri" might be a good choice.

- https://fonts.google.com/specimen/Amiri

This font will show Arabic text slightly larger than what we currently see on the DLME site, and look better/more consistent when there is mixed Arabic and Latin text (as is the case for metadata values for some records).

There might be a lot more to this when we get into the details, but as a start:

- [ ] Embed the Amiri font into `<head>` so all users will have the font
- [ ] Specify a `font-family: 'Amiri', serif;` selector in the CSS so that when the locale is "Arabic" we use that font-family stack

A couple of things we might do to follow this up:

- Figure out how to use the Amiri font for metadata fields (or show page item titles) we know are displaying Arabic, even when in the English locale
- Change the default font stack (for English locale) to use a font that might be more complementary to Amiri

mejackreed:
## Why was this change made?
This change fixes #655 but in a way not originally specified.

### Why?
Great so just load a specific font for a language selector, right this is easy?
https://www.w3.org/International/questions/qa-css-lang

Nope.. so that doesn't work because we need to be more specific. We aren't adding in `lang="ar"` wherever we display different language text.. but we could. But that might be hard.

### Next?
Ok, so we can load fonts for different [unicode ranges](https://developer.mozilla.org/en-US/docs/Web/CSS/%40font-face/unicode-range). And this would be exactly what we want to enable a custom font for [Arabic unicode character ranges](https://en.wikipedia.org/wiki/Arabic_script_in_Unicode). Unfortunately it seems like browser implementation of how unicode range fallbacks work are different (as much as I can tell). Chrome/Safari, everything is good. Firefox no dice, things fallback to the [generic `font-family`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family) name.

## Solution!
Evaluating several Arabic fonts from Google Web Fonts, I came across [Mada](https://fonts.google.com/specimen/Mada?selection.family=Mada). Mada has both Latin and Arabic glyph support. This gives us the exact behavior we are looking for by having a Latin and Arabic styled font. So going forward.. Mada is our font for everything in DLME. We can add different weights in the future if we need them (for Latin only).

I paired w/ @ggeisler on this and got a 👍 

Screen Shot 2019-11-19 at 4 29 30 PM
Screen Shot 2019-11-19 at 4 16 12 PM

Nope.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,6 +1,7 @@
 @import 'variables';
 @import 'spotlight/variables_bootstrap';
 @import 'bootstrap';
+@import 'bootstrap-overrides';
 @import 'blacklight';
 @import 'blacklight_gallery';
 @import 'blacklight_hierarchy';
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -1,0 +1,3 @@
+@import url('https://fonts.googleapis.com/css?family=Mada&display=swap');
+
+$font-family-base: 'Mada', $font-family-sans-serif;


app/assets/stylesheets/application.scss
```diff
@@ -1,6 +1,7 @@
 @import 'variables';
 @import 'spotlight/variables_bootstrap';
 @import 'bootstrap';
+@import 'bootstrap-overrides';
```
cbeer: Could these just go into the current variables file?

mejackreed: No. Because of the import order. We could look at modifying the bootstrap import first, but this felt like the logical way forward.

PR #727, opened 2019-11-21
Remove middleware that set the Content-Type for the SNS HTTP…

jkeck:
…Subscription (and parse raw_post directly).

## Why was this change made?
This was causing issues where every error report's stack trace would indicate that the error occurred in the middleware.

In code comment added/moved from the middleware code


app/controllers/transform_results_controller.rb
```diff
@@ -26,7 +26,11 @@ def show
 
   # @return [JSON] the deserialized JSON within the ++Message++ attribute of the JSON request body
   def notification_msg
-    @notification_msg ||= JSON.parse(params['Message'])
+    # We need to parse the raw_post data because AWS sends this to us as
+    # 'Content-type: text/plain' and rails discards it as part of the params hash.
+    # We then parse the 'Message' payload of the JSON which is also a JSON serialized string.
+    # See https://forums.aws.amazon.com/thread.jspa?threadID=69413
+    @notification_msg ||= JSON.parse(JSON.parse(request.raw_post)['Message'])
```
cbeer: Could we extract `JSON.parse(request.raw_post)` to its own method? The double JSON parse makes my head hurt.

PR #730, opened 2019-11-21
Install Webpacker + Mirador 3

### LINKED ISSUE #726: Should we setup DLME with webpacker to support the M3 usage?
@cbeer Don't want to jump the gun on this, but with the M3 tickets coming on board do we want to go ahead and get that setup?

Just noting that their may be some complexities w/ Spotlight + Docker stuff.  🤷‍♂

camillevilla:
Mirador 3 in a show page

## Why was this change made?
Fixes #726  

### Considerations
- This PR uses the Mirador 3 UMD build, so it isn't IE / Edge compatible 🥺
- Unfortunately, this means that we are using both Sprockets AND webpacker in DLME. 

### Outstanding work:
- Move case statement (`#iiif_manifest_url`) for mapping institutions to IIIF manifest URL patterns upstream to indexing (#734)
- Use Mirador 3 ES build for IE / Edge compatibility (#733)
- Load Spotlight's other JS dependencies via webpacker so we're not using Sprockets AND webpacker (🎟)

- `yarn install` was added to the standard setup script and the Circle CI config.


.circleci/config.yml
```diff
@@ -42,6 +42,10 @@ jobs:
         name: Bundle Install
         command: bundle check || bundle install
 
+    - run:
```
cbeer: Should we be caching the yarn stuff as well?

app/javascript/packs/mirador.js
```diff
@@ -0,0 +1,43 @@
...
+
+
+// Uncomment to copy all static images under ../images to the output folder and reference
+// them with the image_pack_tag helper in views (e.g <%= image_pack_tag 'rails.png' %>)
+// or the `imagePath` JavaScript helper below.
+//
+// const images = require.context('../images', true)
+// const imagePath = (name) => images(name, true)
+
+// import Mirador from 'mirador/dist/es/src/index.js'
```
cbeer: Is this dead code?

app/models/solr_document.rb
```diff
@@ -35,6 +35,19 @@ def embeddable?(blacklight_config = CatalogController.blacklight_config)
...
 
+  def iiifable?
+    iiif_manifest_url.present?
+  end
+
+  def iiif_manifest_url
+    case first('agg_provider.en_ssim')
+    when 'Bodleian Libraries'
+      "https://iiif.bodleian.ox.ac.uk/iiif/manifest/#{first('agg_is_shown_at_ssim')&.split('/')&.last}.json"
+    else
```
cbeer: Should we just omit the whole `else` block? But, 🤷‍♂ , hopefully this whole method isn't long for this world.

app/javascript/packs/mirador.js
mejackreed: Yep.. but what we will want for #733

bin/setup
```diff
@@ -19,8 +19,8 @@ chdir APP_ROOT do
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
 
-  # Install JavaScript dependencies if using Yarn
-  # system('bin/yarn')
+  puts "\n== Install JavaScript dependencies =="
```
cbeer: 👏

app/views/catalog/_show_with_viewer.html.erb
```diff
@@ -1,4 +1,7 @@
-<% if document.embeddable? %>
+<% if document.iiifable? %>
```
cbeer: Leaving this for @ggeisler ; what's the relative priority of IIIFable stuff we display in a DLME-specific mirador viewer vs oembeddable resources to display in a partner-provided viewer (e.g. sul-embed, or A/V, 3D, or PDF viewers)?

.circleci/config.yml
camillevilla: Sure. I'll submit a follow-up PR.

PR #736, opened 2019-11-22
Use an array of field names for the index title configuration.

jkeck:
Spotlight's `Search` model uses blacklight field configurations and injects fields into the `fl` param. 
https://github.com/projectblacklight/spotlight/blob/a8c60279a9c11821e2c262935802d796ab0adfd8/app/models/spotlight/search.rb#L74-L80

Since we set `index.title_field` to a `Blacklight::Configuration::Field` object this causes solr to throw a 400 error.

This may be something we should ultimately fix up-stream, but we can fix it locally here by doing this (although I don't know if this is what we would actually want)

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -70,7 +70,12 @@ class CatalogController < ApplicationController
     end
 
     multilingual_locale_aware_field.call('cho_title') do |field_config|
-      config.index.title_field = Blacklight::Configuration::Field.new(first: true, **field_config)
+      title_field_pattern = Blacklight::Configuration::Field.new(first: true, **field_config)[:pattern]
+      pref_langs, options = lang_config[I18n.locale]
+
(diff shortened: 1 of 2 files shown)


cbeer:
I thought we went away from this approach in order to support the multilingual fallback strategy?

jkeck:
Yeah, agree, that's why I say that I don't think this is actually what we want (and am keeping this as a draft).  It just illustrates the issue (and shows the cause).

Although, when thinking through this, the multilingual fallback strategy is definitely important for most fields, but when for a field where we'll only show one value (like the title) is picking the first available field from a priority ordered list of fields going to end up doing the right thing?  I'm probably not thinking through this enough given our real world data.

jkeck:
Outside of that, I'll look at pushing something up to Spotlight to sanitize that array of field names before sending to solr (assuming we're doing that `fl` for a good reason in the first place).

cbeer:
👍 ; I wonder if we should just have (or support) an explicit query parameter configuration and let applications do whatever they need to.

jkeck:
Fixed by projectblacklight/spotlight#2311

PR #747, opened 2019-11-25
Use the IIIF manifest field from the IR to determine if a document is IIIF-able

cbeer:
## Why was this change made?

The IR should tell us what the IIIF manifest for a resources. Unfortunately, it doesn't seem like any currently indexed data has this information.


camillevilla:
We left this PR unmerged so that we could demo the Mirador viewer at demo 12/3. We could merge this today to help Infra verify that IIIF manifest URLs are being indexed to the correct field.

aaron-collier:
Agree with @camillevilla - I think we're good to merge and retransform as needed. Looking at the fields used I believe we are already setting those fields (when we have them) and will go through the configs and see if we're missing anything.

PR #768, opened 2019-12-03
Add new language codes

aaron-collier:
## Why was this change made?

This adds incoming language codes from dlme-transform in order to keep DLME and DLME-Transform inline with each other for allowed languages.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -36,4 +36,8 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
-  - tr
+  - ms
+  - tr-Arab
+  - tr-Latn
+  - ur-Arab
+  - ur-Latn


mjgiarlo:
Defer to @jacobthill ultimately.

config/settings.yml
```diff
@@ -36,4 +36,9 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
+  - ms
   - tr
+  - tr-Arab
```
cbeer: I thought we /just/ removed `tr-Arab` and `tr-Latn`?

mjgiarlo: @jacobthill ?

jacobthill: We did just remove those codes in favor of 'tr' because nobody writes Turkish in Arabic script (Ottoman Turkish) anymore so I assumed, incorrectly, that we wouldn't have metadata in Ottoman Turkish. We do, however, because the title and creator field sometimes come in that form. My apologies, but we will need them after all.

mjgiarlo:
Apologies if I approved this PR out of turn. Should we ask someone on @sul-dlss/access-team to approve before merging it?

config/settings.yml
cbeer: Thanks @jacobthill; are you saying `tr` and `tr-Latn` are effectively equivalent, then?

jacobthill: Yes, I was assuming only Turkish in Latin script so everything currently using `tr` is actually `tr-Latn`

cbeer: And we'll remediate the things that are currently `tr` to actually be `tr-Latn` and then remove `tr`, or is there some compelling reason to keep both despite the potential for confusion?

jacobthill: No need to keep `tr`, these can all be changed to `tr-Latn`

config/settings.yml
```diff
@@ -36,4 +36,9 @@ acceptable_bcp47_codes:
   - fa-Arab
   - fa-Latn
   - fr
+  - ms
   - tr
```
cbeer:
```suggestion
```

aaron-collier: Removed. I didn't want to remove it if it caused issues with existing data, but it's gone :-)

PR #776, opened 2019-12-05
Remove the browser default padding for facet hierarchies

### LINKED ISSUE #774: Fix alignment of subtypes in hierarchical facets
Screen Shot 2019-12-04 at 4 19 49 PM
Subtypes are pushed to the left and it appears that there are five subtypes instead of three.

cbeer:
After:
Screen Shot 2019-12-05 at 06 57 52

Screen Shot 2019-12-05 at 07 23 57

## Why was this change made?

Fixes #774?


ggeisler:
This looks good to me, and was what I was thinking would solve #774, which it looks like it does.

PR #778, opened 2019-12-05
fixes issue where blank lines in the data cause a delete failure

aaron-collier:
## Why was this change made?

if a blank line exists in the metadata, the delete resources job will fail.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,7 +6,7 @@ require:
   - rubocop-rspec
 
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 2.5
   DisplayCopNames: true
   Exclude:
     - 'bin/webpack*'
--- a/app/jobs/delete_resources_job.rb
+++ b/app/jobs/delete_resources_job.rb
@@ -7,19 +7,17 @@ class DeleteResourcesJob < ApplicationJob
   def perform(url, exhibit)
     resp = Faraday.get(url)
 
-    resource = resp.body.split("\n")
-    resource.each do |json|
-      remove_resource(json: json, exhibit: exhibit)
+    resources = NdjsonNormalizer.normalize(resp.body, url)
+    resources.each do |item|
+      remove_resource(item: item, exhibit: exhibit)
     end
-    logger.info("#{resource.count} records were removed from #{url}.")
+
+    logger.info("#{resources.count} records were removed from #{url}.")
   end
 
   private
 
-  def remove_resource(json:, exhibit:)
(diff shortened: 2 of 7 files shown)


app/jobs/delete_resources_job.rb
```diff
@@ -17,6 +17,8 @@ def perform(url, exhibit)
   private
 
   def remove_resource(json:, exhibit:)
+    return if json.empty?
```
mjgiarlo: I wonder if you'd be open to removing empty lines in the body-parsing code on line 10:

```ruby
resource = resp.body.split("\n").reject(&:empty?)
```

That would prevent having to check later.

.rubocop.yml
```diff
@@ -6,7 +6,7 @@ require:
   - rubocop-rspec
 
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 2.5
```
mjgiarlo: Needed to bump this to allow block-level `rescue` [here](https://github.com/sul-dlss/dlme/pull/778/files#diff-f9ea5052e1fd801bed3e93c8f3fe10d6R27)

mjgiarlo:
@ndushay :speech_balloon: 
> I don't fully understand the issue here -- is the incoming JSON actually non-compliant and using newlines instead of commas? Because each "item" is just a line in the file? that's annoying.

According to the NDJSON spec, every line ought to be valid JSON. [Section 3.2 of the spec](https://github.com/ndjson/ndjson-spec#32-parsing), on the other hand, recommends that parsers optionally silently ignore blank/empty lines. So I'm not sure the incoming JSON is *bad*, exactly, but our code in DLME was previously assuming every line is valid JSON, including blank lines.

This PR takes the code that splits and parses the NDJSON content and concentrates it into its own class so we're not doing empty-checks all over the place.

bin/yarn
```diff
@@ -3,11 +3,9 @@
...
 APP_ROOT = File.expand_path('..', __dir__)
 Dir.chdir(APP_ROOT) do
-  begin
-    exec 'yarnpkg', *ARGV
-  rescue Errno::ENOENT
-    $stderr.puts 'Yarn executable was not detected in the system.'
-    $stderr.puts 'Download Yarn at https://yarnpkg.com/en/docs/install'
-    exit 1
-  end
+  exec 'yarnpkg', *ARGV
```
mjgiarlo: This is a rubocop-related change, now that we're allowing Ruby 2.5+ syntax (including block-level `rescue`) in its checks.

PR #787, opened 2019-12-09
Display range limit as ltr when in rtl Arabic view fixes #777

### LINKED ISSUE #777: Style issues with Arabic date slider view
Right to left conversion issues in date range slider.

- minus sign is to the right of the number instead of left
- numbers pushed off the screen 

Could possibly be solved by using the English slider since numbers go left to right in Arabic.

Screen Shot 2019-12-05 at 2 25 44 PM

mejackreed:
## Why was this change made?
To fix #777 

Screen Shot 2019-12-09 at 3 27 57 PM
Screen Shot 2019-12-09 at 3 27 47 PM

### DIFF
--- a/app/assets/stylesheets/rtl.scss
+++ b/app/assets/stylesheets/rtl.scss
@@ -124,4 +124,10 @@ $breadcrumb-item-padding-x: 0.5rem !default;
       padding-right: 0;
     }
   }
+
+  .limit_content.range_limit{
+    .profile, form.range_limit {
+      direction: ltr;
+    }
+  }
 }


cbeer:
I know this is what was asked for, but... shouldn't the "Unknown" and count display RTL-style (and "view larger", presumably?)

mejackreed:
I guess? 🤷‍♂ . I've updated this and the screenshots. Is this what you mean?

ggeisler:
> shouldn't the "Unknown" and count display RTL-style (and "view larger", presumably?)

It would be great to show those RTL. I just wasn't sure how hard it would be to do those separately from the rest of the panel contents so didn't consider it a requirement.

mejackreed:
Cool.. then I think the PR with the update is correct.

PR #801, opened 2019-12-10
Replace the Blacklight logo with a DLME text logo

ggeisler:
## Why was this change made?
First step in visual design. A DLME image logo doesn't currently exist (I believe Wayne said they hoped to design one at some point but it is not work that is currently in plan) but we do need to replace the Blacklight logo in the topbar with something more specific to DLME.

At least for now, I simply replaced the BL logo with some text, removed the link, and tightened the height of the topbar a bit. Some of this might be adjusted as we get further along with the visual design (and if we think of a better way to handle the DLME logo).

Screen Shot 2019-12-10 at 4 30 11 PM

---
I don't think there is (or should be?) an Arabic translation for "DLME" so nothing needs adjusting for RTL, unless @jacobthill thinks it would make sense to translate "DLME".

N/A


app/views/shared/_header_navbar.html.erb
```diff
@@ -0,0 +1,12 @@
+<nav class="navbar navbar-expand-md navbar-dark bg-dark topbar" role="navigation">
+  <div class="<%= container_classes %>">
+    <div class="project-logo">DLME</div>
+    <button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#user-util-collapse" aria-controls="user-util-collapse" aria-expanded="false" aria-label="Toggle navigation">
```
camillevilla: @ggeisler Can we use the i18n `aria-label` here, similar to the [upstream partial](https://github.com/projectblacklight/spotlight/blob/master/app/views/shared/_header_navbar.html.erb)? 

We don't have this key in Spotlight's `ar.yml` yet, but hopefully we'll have it at some point in the future.
```suggestion
    <button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#user-util-collapse" aria-controls="user-util-collapse" aria-expanded="false" aria-label="<%= t :toggle_nav %>">
```

ggeisler: Sure, good eye @camillevilla. I started with the Blacklight version of this partial; I probably should've started with the Spotlight version.

PR #806, opened 2019-12-11
Add Arabic DLME-specific locale, fixes #766

### LINKED ISSUE #766: provide DLME-specific locale file translation
@jacobthill Translations for these keys will be need to provided in an `ar.yml` file. Note that a good chunk of the items in [DLME's en.yml](https://github.com/sul-dlss/dlme/blob/master/config/locales/en.yml) do not need to be translated because they are curator-facing (i.e. "Add DLME JSON file").

https://github.com/sul-dlss/dlme/blob/66f1558f348533f1d616e0f70b9f990be53730e7/config/locales/en.yml#L47-L72

jacobthill:
## Why was this change made?

- Add the Arabic DLME-specific locale file to enable translation of labels that are specific to DLME 

- n/a


cbeer:
Thanks @jacobthill ; I think we'd  prefer to omit untranslated labels entirely so they fall back to the English labels instead of (re-)supplying them in the Arabic translations.

PR #819, opened 2019-12-13
Update styling for RTL facet remove icon display fixes #788

### LINKED ISSUE #788: RTL facet remove icon display
The "remove facet" icon is misplaced when the value is, itself, RTL and needs a little more padding before the value

Screen Shot 2019-12-10 at 07 49 37

In RTL  mode with LTR values, the icon is misplaced to the right of the value:

Screen Shot 2019-12-10 at 07 55 18

In LTR mode, it looks more like:

Screen Shot 2019-12-10 at 07 55 30

mejackreed:
Screen Shot 2019-12-13 at 1 30 11 PM
Screen Shot 2019-12-13 at 1 29 48 PM

### DIFF
--- a/app/assets/stylesheets/rtl.scss
+++ b/app/assets/stylesheets/rtl.scss
@@ -1,6 +1,8 @@
 $breadcrumb-item-padding-x: 0.5rem !default;
 @import 'bootstrap/rtl';
 
+.facet-label .facet-select,
+.facet-label .selected,
 .filter-value,
 .tooltip-inner,
 .document-metadata dd,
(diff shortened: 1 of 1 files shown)


cbeer:
We need to keep `unicode-bidi: plaintext` for `.facet-label`. Without it, you get things like:

Screen Shot 2019-12-13 at 13 28 56

mejackreed:
@cbeer got it.. what about something like this a8024c7c554cb83099091fbc24eb7c24591da63b

PR #826, opened 2019-12-17
Setup contextual alert dismissable for search results fixes #805

### LINKED ISSUE #805: Contextualize data provided by third parties so it's more clear why data in other languages may be displaying
It is confusing to have records with data elements in English only. Why would there be an Arabic version of the site if there are items that only have English titles -- ie, they can't be searched in Arabic? This feels confusing and problematic, and would benefit from some user-facing explanation.

Part of https://github.com/sul-dlss/dlme/issues/800

mejackreed:
## Why was this change made?
To fix #805

alert

## Unanswered question
I'm not sure how we are going to provide a link within this alert to a user created about page. Any thoughts?

### DIFF
--- a/app/controllers/session_displays_controller.rb
+++ b/app/controllers/session_displays_controller.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+##
+# A controller for disabling session information
+class SessionDisplaysController < ApplicationController
+  ##
+  # Update the session, to disable_search_context
+  def update
+    session[:disable_search_context] = true
+    redirect_to request.referer
+  end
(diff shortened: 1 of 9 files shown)


jkeck:
WRT how we link to the about page, although it's hokey, we could have a named route match a `GET` to `/library/about/searching-this-site` (which should be consistent across all environments), or in that case just use `exhibit_about_page_path(current_exhibit, 'searching-this-site')`.

We could even make the slug a configurable setting in case we want to / decide to change it later (although I don't know what the likelihood of that is, but could be useful for local development)

PR #888, opened 2020-03-31
Authorize range limit actions from exhibit home pages

### LINKED ISSUE #850: Range limit doesn't load data on the home page for non-logged in users

cbeer:
Fixes #850

### DIFF
--- a/app/controllers/spotlight/home_pages_controller.rb
+++ b/app/controllers/spotlight/home_pages_controller.rb
@@ -10,5 +10,15 @@ module Spotlight
   # Override the upstream HomePagesController in order to inject range limit behaviors
   class HomePagesController
     include BlacklightRangeLimit::ControllerOverride
+
+    # rubocop:disable Rails/LexicallyScopedActionFilter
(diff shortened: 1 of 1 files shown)


app/controllers/spotlight/home_pages_controller.rb
```diff
@@ -10,5 +10,15 @@ module Spotlight
   # Override the upstream HomePagesController in order to inject range limit behaviors
   class HomePagesController
     include BlacklightRangeLimit::ControllerOverride
+    
+    # rubocop:disable Rails/LexicallyScopedActionFilter
+    # Tweak the authorization for the range limit actions
+    before_action :authenticate_user!, except: [:show, :range_limit, :range_limit_panel]
```
jkeck: Should we still be authenticating the user in the `show` action? Or since home pages can't be published/unpublished does the exhibit's accessibility take precedence and it doesn't really end up mattering?

cbeer: I think that's probably right. This is copied straight for exhibits where we solved the same problem.

Issue #765, opened 2019-12-03
Update date range slider styling


ggeisler:
- Hide the chart range handles
- Reduce font size of chart x-axis labels to `14px`

![Screen Shot 2019-12-03 at 12 38 06 PM](https://user-images.githubusercontent.com/101482/70087903-033b0c80-15ca-11ea-9572-f7871e7aaf65.png)

- Reduce the `margin-top` on `.limit_content .subsection` to `0.5em;` from `1em`

![Screen Shot 2019-12-03 at 12 40 15 PM](https://user-images.githubusercontent.com/101482/70087958-22399e80-15ca-11ea-896b-db37e26e8a20.png)

- ~~Widen the chart **and** slider bar horizontally so that it uses the full-width of the `card-body`. In the screenshot below notice how the "Unknown 31,593" label goes to the right and left edges of the card body but the date range chart and slider do not. The extra width would be useful when using the sidebar version of the slider.~~ This has been fixed.

![Screen Shot 2019-12-03 at 11 36 27 AM](https://user-images.githubusercontent.com/101482/70083596-8d32a780-15c1-11ea-8ac5-e4d9ef9ec83a.png)

jkeck:
Submitted projectblacklight/blacklight_range_limit#123 for most of the tasks here.
Submitted a PR upstream to flot to deal w/ the last flot/flot#1726 (removing the handles)

jkeck:
PRs that will unblock the config change that will address the final pieces
- [ ] https://github.com/projectblacklight/blacklight_range_limit/pull/144
- [ ] https://github.com/projectblacklight/blacklight_range_limit/pull/145

jkeck:
I believe this can be closed now.  Feel free to re-open / file new issues for additional work.

Issue #895, opened 2020-04-01
Update how date range slider displays Hijri dates


ggeisler:
I'm creating a new ticket for this to clearly separate this issue from:

- General styling/display issues with the date range slider (covered in #765)
- Display of Hijri dates in metadata (covered in #845)

As noted in #845, the concern is that (if I understand correctly) the Hijri calendar starts with year 1, so displaying a value like "-500" as a label on the x-axis of the date range graph and as a value in the date range inputs doesn't really make sense.

### Preferred solution
It's not clear to @jacobthill and I what is technically feasible but here is our preferred solution:

1. For negative year values displayed on the x-axis, drop the minus sign and append "BH" to the year
2. For positive year values displayed on the x-axis, append "H" to the year
3. Do the same as 1 and 2 above for years displayed in the input fields
4. Do the same as 1 and 2 above for years displayed in the active constraint label

<img width="642" alt="71283802-381bc300-232f-11ea-918a-59635810489e" src="https://user-images.githubusercontent.com/101482/78194102-0be50780-7431-11ea-90e7-867ebe9da3ef.png">

(Ignore the flash message shown next to the facet title; this would be very challenging to display nicely in the sidebar version of the facet, and I'm hoping it is not really necessary.)

 ### Potential lesser solution
This is not at all ideal, but I'm worried that the input fields (item 3 above) might be more challenging than the other items, because it needs to be strictly numeric. If that's true, we might be able to get away with just doing:

1. For negative year values displayed on the x-axis, drop the minus sign and append "BH" to the year
2. For positive year values displayed on the x-axis, append "H" to the year
3. Do the same as 1 and 2 above for years displayed in the active constraint label

and leaving the input fields displaying a minus sign, to keep the field numeric-only. It would be very awkward to use different formatting for the input fields versus the labels, but perhaps it will nonetheless be clear to the user (because the numeric values will still match) that "-500" in the input box is shown as "500 BH" on the x-axis. 

I really don't like this solution, but bring it up as possibly better than doing nothing at all, if we can't do all of the preferred solution.

cbeer:
If we do the preferred solution, how should we handle user input? Will they be typing H + BH (and, what variants of that do we need to support?)?

jacobthill:
I think we should opt for the "potential lesser solution". I think having users type letters will be just trading one UX problem for another. This is not perfect but it is an improvement. I think we should also do the same thing to the Gregorian calendar view for consistency.

[...]

PR #896, opened 2020-04-02
Update the range limit plugin and configure

jkeck:
## Why was this change made?
To wrap up work on #765, primarily configuring flot to no longer display the selection decorations/handles on the year histogram.

n/a

## Before
facets-before

## After
facets-after

### DIFF
--- a/app/assets/javascripts/range_limit_config.js
+++ b/app/assets/javascripts/range_limit_config.js
@@ -1,0 +1,12 @@
+Blacklight.onLoad(function() {
+  var dateRangeFacets = [
+    $('.blacklight-cho_date_range_norm_isim'),
+    $('.blacklight-cho_date_range_hijri_isim')
+  ];
+
+  for(i = 0; i < dateRangeFacets.length; i++) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/range_limit_config.js
```diff
@@ -0,0 +1,12 @@
+Blacklight.onLoad(function() {
+  var dateRangeFacets = [
+    $('.blacklight-cho_date_range_norm_isim'),
+    $('.blacklight-cho_date_range_hijri_isim')
+  ];
+
+  for(i = 0; i < dateRangeFacets.length; i++) {
+    dateRangeFacets[i].data('plot-config', {
```
mejackreed, line 8: I'm thinking we might want to do something along this line https://github.com/flot/flot/blob/master/API.md#customizing-the-legend to also support #895 ?

jkeck: I'm not sure I understand the ask.

mejackreed: Not an ask... just started thinking about that future ticket.

PR #897, opened 2020-04-02
Use the Markazi Text font 

### LINKED ISSUE #881: Find new Arabic font
Several data partners noted that the Arabic font chosen for DLME looks rather old fashioned and some letters are difficult to read. Can we find a more modern/legible font?

cbeer:
Screen Shot 2020-04-02 at 13 02 40
Fixes #881


ggeisler:
Thanks for creating this branch so I could test things out @cbeer.

Out of the box Markazi Text does not look too good in English:

Screen Shot 2020-04-02 at 3 21 46 PM

As I think is pretty clear in the screenshot above (even easier to see if you compare side-by-side to the current homepage at https://spotlight.dev.dlmenetwork.org/library), it's a very small font, quite dark, and seems to have a noticeable amount of space above and below the actually font itself (don't know the official term for this, but it's like it has built-in additional line-height). I'm not an expert on typography at all, so maybe these are naive criticisms, but at least with our current typographic styling the Latin version seems kind of weird and unacceptable as-is.

However, if we're willing to invest some time into customizing our typography-related CSS for this font, I think we can make it work. As long as the typography updates we need to make for the English to look good don't detract from the Arabic version.

For example, just to get a sense of what we'd have to do to improve the English version of the site if we went with this font, below I increased the body text, facet heading, and site title heading sizes, and reduced the font-weight and line-height of the body text and facet headings:

Screen Shot 2020-04-02 at 3 18 13 PM

We'd have to make these sort of updates throughout but it seems feasible to get it looking decent in English. With those same changes, here's how things look when you switch to the Arabic version:

Screen Shot 2020-04-02 at 3 18 27 PM

@jacobthill Before we commit to this font update, can you have QNL take a look at the two screenshots above and make sure that they are okay with them, with the assumption that these roughly illustrate how things will turn out if we switch to Markazi Text?

jacobthill:
I sent this along to QNL.

anarchivist:
@cbeer @ggeisler Would a technique like using the [`:lang()` CSS pseudoselector](https://www.w3.org/International/questions/qa-css-lang.en) work to apply this just to the Arabic text?

ggeisler:
@anarchivist Yeah, I definitely thought about a backup plan where we can use the `lang` value to use different CSS for each language. For instance, if we hear from QNL that the CSS updates we need to make for English (e.g., increasing font size and decreasing line-height) that would be an approach I hope we can use.

But I'm hoping those sort of changes don't noticeably detract from the Arabic version.

mejackreed:
@anarchivist @ggeisler I _think_ I tried that in https://github.com/sul-dlss/dlme/pull/709. But what I ran into was that we can't and don't always specify the `lang` attributes. Which I think without this gives you a "mixed" font look. We could try this again just to confirm ?

jacobthill:
I think the impact to the Arabic font is minimal and is a slight improvement in my opinion. I doubt QNL will see any problems with it so I wouldn't put any effort into it unless we hear from QNL that there is a problem.

anarchivist:
Thanks @mejackreed - it looks like we're declaring the `lang` attribute on the root `<html/>` element through internationalization.

I think this will possibly still have some impacts on things like the metadata fields - like the titles, which might be in a language other than Arabic.

mejackreed:
When using the `lang` selector, you do get a disconnect between these. I think this might impact spacing differently as well and we might have to account for that. I think maybe that is why I offered using a font with both Arabic and Latin glyphs was important. (Just trying to remember the edge cases now). Maybe that is fine?

## AR
Screen Shot 2020-04-03 at 11 25 52 AM

## EN
Screen Shot 2020-04-03 at 11 25 41 AM

PR #899, opened 2020-04-02
Allow contributing institutions to exist in more than one country.

### LINKED ISSUE #883: Agg providers with multiple countries only display one
American Institute for Maghrib Studies has three countries in the `agg_data_provider_country` field but only one shows up in the statistics dashboard. We need all three to display.
Screen Shot 2020-03-30 at 2 07 03 PM
Screen Shot 2020-03-30 at 1 56 26 PM

jkeck:
Closes #883 

## Why was this change made?
Institutions can exist in multiple countries so we need to display all of them in the dashboard (we were using just the first returned in the pivot facet response for display).

n/a

Using the AIMS transform from dev on 3/12/2020

## Before
contributors-before

## After
contributors-after

### DIFF
--- a/app/models/statistics_dashboard.rb
+++ b/app/models/statistics_dashboard.rb
@@ -163,7 +163,7 @@ class StatisticsDashboard
     end
 
     def total_countries
-      institutions.collect(&:country).uniq.count
+      institutions.collect(&:countries).flatten.uniq.count
     end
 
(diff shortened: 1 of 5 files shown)


ggeisler:
Looks good @jkeck. I believe we also wanted to change the column heading to "Countries" plural.

PR #902, opened 2020-04-03
Display range chart with BH/H BCE/CE labels

### LINKED ISSUE #895: Update how date range slider displays Hijri dates

mejackreed:
Fixes #895 
Adds BCE/CE and BH/H suffixes to range limit and constraints in both English and Arabic. Confirmed correct behavior on consultation with @ggeisler and @jacobthill .

*Note* the behavior in the screenshot represents integration with https://github.com/projectblacklight/blacklight_range_limit/pull/148. This needs to be released and shipped (or similar) for everything to work properly.

## Updated screen shots
Screen Shot 2020-04-07 at 13 38 01
Screen Shot 2020-04-07 at 13 37 17

## Old Screenshot
Screen Shot 2020-04-03 at 9 58 52 AM


ggeisler:
@mejackreed Great to see we're able to add the date abbreviations to the numeric value. I realize that this is really challenging our horizontal space but I think we might need to include a space between the date's numeric value and the "BH" or "H" suffix. 

Can we try that and see how it looks?

mejackreed:
Sure thing. Here is an updated shot with the spaces.
Screen Shot 2020-04-03 at 10 20 35 AM

ggeisler:
👍 Thanks, I think we should go with the version with the space.

app/assets/javascripts/range_limit_config.js
```diff
@@ -1,12 +1,33 @@
...
+  });
+  
+  var gregorianConfig = $.extend({}, config, {
+    xaxis: {
+      tickFormatter: genericTickFormatter(['bce', 'ce'])
+    }
+  })
+
+  $('.blacklight-cho_date_range_norm_isim').data('plot-config', gregorianConfig);
+  $('.blacklight-cho_date_range_hijri_isim').data('plot-config', hijriConfig);
```
mejackreed, line 32: Once we get a new release of BLRL, we can extend this to also do this on modal open.

app/helpers/override_range_limit_helper.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# Override range limit provided helpers
+module OverrideRangeLimitHelper
+  include RangeLimitHelper
+
+  ##
+  # Overriden helper to add the BH/H and BCE/CE to labels
+  def range_display(solr_field, myparams = params)
+    display = super(solr_field, myparams)
```
jkeck, line 10: I _believe_ `super` w/o args passes through the args methods to the super method.

mejackreed: Oh good call. I've updated accordingly.

PR #929, opened 2020-04-07
Remove field-specific configuration from the statistics dashboard query

### LINKED ISSUE #928: Wrong number of contributors in statistics page
Screen Shot 2020-04-07 at 1 08 59 PM

I'm not clear where the contributors are coming from but my understanding is that this should be dynamically updated when new data is loaded. There should be more contributors. `Providers (Indirect)` has 22 institutions and `Providers (Direct)` has 20.

cbeer:
Fixes #928

### DIFF
--- a/app/controllers/statistics_controller.rb
+++ b/app/controllers/statistics_controller.rb
@@ -8,6 +8,12 @@ class StatisticsController < Spotlight::ApplicationController
   before_action :attach_breadcrumbs
   before_action do
     authorize!(:read, current_exhibit)
+
+    blacklight_config.facet_fields.clear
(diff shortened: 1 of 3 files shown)


app/controllers/statistics_controller.rb
```diff
@@ -8,6 +8,12 @@ class StatisticsController < Spotlight::ApplicationController
   before_action :attach_breadcrumbs
   before_action do
     authorize!(:read, current_exhibit)
+
```
cbeer, line 11: I toyed around with explicitly setting (infinite) limits for the fields we care about instead. I could go either way.

Issue #1003, opened 2020-05-21
Statistics page displays wrong data for Item contributors


jacobthill:
`Item contributor` should pull from `agg_data_provider`.  This was probably addressed by https://github.com/sul-dlss/dlme/issues/936 but the statistics page is currently displaying the wrong country values for two orgs: 

<img width="1124" alt="Screen Shot 2020-05-21 at 9 46 41 AM" src="https://user-images.githubusercontent.com/37662787/82565473-7caec300-9b48-11ea-8f5e-ccd7b86bfe2d.png">

Both institutions should have France for `Data contributor` and Egypt for `Item contributor`. The IR has the correct data. Do these records need to be reloaded or reindexed for changes to take effect or did the change in https://github.com/sul-dlss/dlme/issues/936 not get implemented correctly?

jacobthill:
It seems the statistics page just has stale data. The production site has some type values in the statistics page that don't show up in the facets on the home page:
<img width="560" alt="Screen Shot 2020-05-21 at 9 58 21 AM" src="https://user-images.githubusercontent.com/37662787/82566386-d6fc5380-9b49-11ea-8366-6a7b480dd8c4.png">

cbeer:
@jacobthill - can you  say more about in what way the data is wrong?

For the items table, it seems like there may be some differences between the fields we're using. Are you asking us to change the statistics page to only pull from the cho_type_facet?

jacobthill:
Sure, I have a meeting in a few minutes but looking quickly `NOT FOUND` and `Oral narrative` show up on the statistics page but not on the facets in the home page. 
<img width="556" alt="Screen Shot 2020-05-21 at 10 47 55 AM" src="https://user-images.githubusercontent.com/37662787/82571446-f8147280-9b50-11ea-815b-dd3f0c8f1a02.png">
I'm not sure if these are the only errors but the statistics page should reflect the same data as the facets. I'm not sure what `cho_type_facet` references. The facet that is displayed on the home page is just called `Type` in the spotlight UI and in the IR this comes from `cho_edm_type` and `cho_has_type`.

cbeer:
It  looks like `cho_type_facet` is calculated  over in dlme-transform using  the same fields, but clearly can get out of sync. Perhaps there's some old data in -prod that needs to be reindexed?

Looking at  some of these records:
https://dlmenetwork.org/library/catalog?f%5Bcho_edm_type.en_ssim%5D%5B%5D=NOT+FOUND

[...]

PR #1004, opened 2020-05-21
Use agg_data_provider_country for item provider stats

cbeer:
## Why was this change made?

Fixes part of #1003 by using the agg_data_provider_country field for data providers

### DIFF
--- a/app/models/statistics_dashboard.rb
+++ b/app/models/statistics_dashboard.rb
@@ -12,12 +12,16 @@ class StatisticsDashboard
     'facet.field': [
       'agg_provider_country.ar-Arab_ssim',
       'agg_provider_country.en_ssim',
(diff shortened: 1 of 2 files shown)


cbeer:
@jacobthill can you confirm that this was the  error you were pointing  out in  #1003, and (@ggeisler too..) that  this is  the desired behavior? AFAICT, we never  talked about what country field  to use when we split the stats page into  two parts.

jacobthill:
I can't tell exactly what effect those changes will have so I can't confirm that it is the error. Maybe it will be more clear to @ggeisler. Essentially what should happen is:

agg_provider >> Data Contributor
agg_provider_country >> Data Contributor country
agg_data_provider >> Item Contributor
agg_data_provider_country >> Item Contributor country

cbeer:
Thanks, that's what I needed to know.

PR #1040, opened 2020-07-23
Update docker build image to handle Ruby and NodeJS version dependenc…

mejackreed:
…y needs

## Why was this change made?
Could not successfully build the docker image locally


jkeck:
Updated in 2fb88e74be94d19c4b587c77e3976c646d795c24

PR #1048, opened 2020-08-10
Update dependencies

sul-devops-team:


Gemfile.lock
```diff
@@ -124,20 +124,22 @@ GEM
...
       openseadragon
       paper_trail (~> 10.0)
-      rails (~> 5.1)
+      rails (< 6.1)
       riiif (~> 2.0)
       roar (~> 1.1)
       roar-rails
       signet
-      sir_trevor_rails (~> 0.5)
+      sir_trevor_rails (>= 0.6.1)
```
jkeck, line 132: I'm a bit surprised to not see the generated block configuration initializer that sir_trevor_rails 0.6 introduced.  My guess is we may not have a test that exercises block creation in DLME.

PR #1100, opened 2021-01-12
Fixes #1074, by proxying http only images and disabling http iiif content

### LINKED ISSUE #1074: Revisit options to deal with http content
American Numismatic Society does not use https which prevents us from displaying images. We have discussed this before with them and they seem reluctant to change this policy. QNL is concerned that the way their content looks in DLME  will reflect poorly on DLME in the eyes of the users. We need to revisit this issue to see if there is another solution. 

The same problem is occurring with CSIC data: https://dlmenetwork.org/library/catalog?f%5Bagg_provider_en%5D%5B%5D=Consejo+Superior+de+Investigaciones+Cient%C3%ADficas+%28CSIC%29

mejackreed:
## Why was this change made?
Fixes #1074

This adds a protected proxy for http images and disables http iiif content which we cannot view.

An additional thing added was the `loading="lazy"` attribute for images so we don't try and proxy non-visible images.

### DIFF
--- a/app/controllers/image_proxy_controller.rb
+++ b/app/controllers/image_proxy_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+##
+# Controller for proxying HTTP images
+class ImageProxyController < ApplicationController
+  before_action :validate_token
+
+  def access
+    send_data proxied_response.body, type: proxied_response.content_type, disposition: 'inline'
+  end
+
+  private
+
+  ##
+  # To prevent outside applications from using this proxy
+  def validate_token
+    render plain: 'invalid token', status: :forbidden unless valid_authenticity_token?(session, image_token)
+  end
+
+  def proxied_response
+    @proxied_response ||= begin
+      benchmark "Fetch #{image_url}" do
+        HTTP.get(image_url)
+      end
+    end
+  end
+
+  def image_url
+    params.require(:url)
+  end
+
+  def image_token
+    params.require(:token)
+  end
+end
--- a/app/controllers/image_proxy_controller.rb
+++ b/app/controllers/image_proxy_controller.rb
@@ -5,8 +5,10 @@
 class ImageProxyController < ApplicationController
   before_action :validate_token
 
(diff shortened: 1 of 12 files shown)


app/controllers/image_proxy_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  ##
+  # To prevent outside applications from using this proxy
+  def validate_token
+    render plain: 'invalid token', status: :forbidden unless valid_authenticity_token?(session, image_token)
+  end
+
+  def proxied_response
+    @proxied_response ||= begin
+      benchmark "Fetch #{image_url}" do
```
mejackreed, line 22: Should we look into Rails.cache here?

app/controllers/image_proxy_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  def access
+    send_data proxied_response.body, type: proxied_response.content_type, disposition: 'inline'
+  end
+
+  private
+
+  ##
+  # To prevent outside applications from using this proxy
+  def validate_token
+    render plain: 'invalid token', status: :forbidden unless valid_authenticity_token?(session, image_token)
```
mejackreed, line 17: Is this not as smart as I think, as are we going to invalidate browser cache on every image?

app/controllers/image_proxy_controller.rb
jkeck: If we throw a rails cache around this would that assuage some of the concern about authenticity tokens breaking browser cache?

mejackreed: I've popped a commit on there that the ability to do this. One drawback is that I don't think we can stream and cache the responses at the same time :(

mejackreed: Ok.. Just added one more.. Using a referer check instead of token based? This will hopefully enable browser caching and still be a useful deterrent, though of course it can be spoofed.

jkeck (approved):

PR #1113, opened 2021-01-14
Fixes #1028 providing a configurable calendar for the date range sliders

### LINKED ISSUE #1028: Convert separate Hijri and Gregorian calendar view to single view with toggle
Screen Shot 2020-06-24 at 8 56 13 PM

The separate calendar views have led to some confusion during our user testing. The behavior was initially confusing and was made more so by https://github.com/sul-dlss/dlme/issues/1027. Some users expected the two calendar views to stay aligned (e.g. sliding one would slide the other). The use case for two separate calendar views is also non-existent; the user would choose one but never need to use both together. I believe it was implemented this way because it was less challenging technically but a single calendar view with a toggle to switch between Hijri and Gregorian would be preferable. This is a request from our colleagues at QNL and the user experience angle makes perfect sense to me. We need to either implement the single calendar with toggle view or, if this is not technically possible, we need to document in a clear and succinct way why it isn't possible so I can communicate this to our colleagues.

mejackreed:
Note: this will temporarily allow us to use the default branch from BlacklightRangeLimit, tracked in a card.

Kapture 2021-01-14 at 09 03 59

I wasn't able to trigger the change events in tests, so we don't have full coverage there, but these do not appear to be tested upstream either.


config/locales/ar.yml
```diff
@@ -19,7 +19,10 @@ ar:
   date_field: "%{gregorian_dates} / %{hijri_dates}"
   date_field_gregorian: "%{dates} (ميلادي)"
   date_field_hijri: "%{dates}(هجري)"
+  date_field_label_gregorian: ميلادي
+  date_field_label_hijri: هجري
   date_range: "%{min} - %{max}"
+  date_range_calendar: Calendar
```
mejackreed, line 25: This has been asked for in the original ticket and is being tracked here: https://github.com/orgs/sul-dlss/projects/35#column-12461630

I'm hoping tracking there should enable us to move forward on review of this.

Gemfile
```diff
@@ -83,7 +83,7 @@ gem 'sitemap_generator'
 gem 'blacklight-gallery', '>= 0.3.0'
 gem 'blacklight-hierarchy'
 gem 'blacklight-oembed', '>= 0.1.0'
-gem 'blacklight_range_limit', '~> 7.4'
+gem 'blacklight_range_limit', github: 'projectblacklight/blacklight_range_limit', branch: 'master'
```
mejackreed, line 86: A new release is being tracked in our backlog. Before doing those mechanics, I wanted to make sure we were good to go.

camillevilla (approved):

PR #1117, opened 2021-01-15
Update Blacklight and Spotlight versions

mejackreed:


cbeer:
I think we want to wait for @jkeck 's blacklight-gallery fix, yes?

jkeck:
Looks like we're using blacklight-gallery 2 here.  Should we update this to BL-Gallery 3.x so that we can get gallery updates, or ship this and follow on later w/ update for blacklight-gallery 3?

mejackreed:
My inclination is to ship this so we can at least get the upstream stuff in for the demo what we have done already. If we do another update w/ dependencies later today or monday.. seems fine w/ me

jkeck (approved):

Issue #639, opened 2019-11-05
Have Arabic numbers display in the Arabic view (e.g. for years)


jacobthill:
Need a way to convert this [1999, 2000, 2001, 2002] into this [١٩٩٩، ٢٠٠٠، ٢٠٠١، ٢٠٠٢] in the application so Arabic numbers can be displayed in the DLME Arabic view.

cbeer:
https://stackoverflow.com/questions/18124562/localization-of-numeric-number-in-rails 🤷‍♂

cbeer:
In what circumstances should we be doing this conversion? Is it:

a) for numbers generated by the UI (facet counts, pagination, etc)
b) for numeric data (say, years extracted from the metadata)
c) any time a number appears in data (which seems potentially fraught and confusing)

ggeisler:
I'll leave it to @jacobthill to provide a more authoritative answer to this question, but I want to suggest a couple of things based on my investigation into RTL and a previous conversation with Jacob.

This ticket is probably the lowest-priority of all RTL-related tickets. As a general rule, readers who prefer Arabic and RTL interfaces do understand and are used to reading _Western Arabic_ numbers (numbers displayed in English and many other languages as: `0, 1, 2, 3, etc.`). So the general goal of this ticket to provide Eastern Arabic translations of numbers is, in my opinion, much less critical than the other RTL issues.

For example, if you do a search in this [Arabic Union Catalog site ](https://www.aruc.org/search) in the Arabic locale (which appears to be the primary locale, with English and French as secondary options), you'll see the pagination still uses Western Arabic numbers. And if you scan through the [SkyNews Arabia site](https://www.skynewsarabia.com/), you'll see plenty of Western Arabic numbers (day of the month, year, etc.)

So my second suggestion is: if and when we do get around to doing number conversion from Western Arabic to Eastern Arabic, it probably makes most sense to limit that to specific metadata fields that we know contain years and only years (e.g. Date, perhaps Date Range). I think this a limited version of Case B from @cbeer's question above.

Case C seems not worth trying at all, and while Case A is probably achievable, since in those instances we know we are only going to encounter numbers, I think we'd want to evaluate whether the benefit is worth the effort, given Arabic readers seem comfortable enough with Western Arabic numbers that at least some Arabic-first sites don't seem to worry about doing the conversion.

jacobthill:
I think we are only interested in b

jacobthill:
[...]

PR #1124, opened 2021-01-16
[WIP] arabic numeral replacement for statistics page

camillevilla:
## Why was this change made?
Part of #639


camillevilla:
Closing in favor of #1165, which does this without wild Rails overrides

PR #1148, opened 2021-01-25
Update to Rails 6.1

cbeer:


app/assets/config/manifest.js
```diff
@@ -1,3 +1,3 @@
 //= link_tree ../images
 //= link_directory ../javascripts .js
-//= link_directory ../stylesheets .css
+//= link application.css
```
mejackreed, line 3: Could this change be the culprit? Upstream, I'm still seeing: https://github.com/rails/rails/blob/v6.1.1/railties/lib/rails/generators/rails/app/templates/app/assets/config/manifest.js.tt

mejackreed (approved):

Issue #146, opened 2017-08-22
Suggest UI design for flagging "bad" records


cbeer:
Perhaps borrowing heavily from Revs?

ggeisler:
Below are mockups for a simple approach, where it doesn't matter if the user is logged in and the user doesn't have the option to find out whether the suggestion she makes is resolved.

(Numbered sections below correspond to the orange numbered icons in the mockups.)

1. Adds  a collapse/expand (e.g., accordion) section above the metadata. Default is closed, so only the section header is shown. 

![flagging-section](https://user-images.githubusercontent.com/101482/29640860-be92f8ca-8815-11e7-9b27-406d4eea018d.png)

2. When the user clicks anywhere in the section header, the section expands to display the form.

![flagging-section-expanded](https://user-images.githubusercontent.com/101482/29641258-9eedc390-8817-11e7-9122-eecf50546354.png)

3. The Submit button sends the form value to the database and/or email address, collapses the section back to the initial state, and displays a flash message to confirm to the user that their suggestion was successfully received:

![flash-message](https://user-images.githubusercontent.com/101482/29641310-d6ef2716-8817-11e7-9598-f754933834ed.png)

4. The Cancel link clears the form input and collapses the section back to the initial state.

Of course, the form could be more sophisticated to help manage incoming suggestions, such as adding a select menu the user would use to add a label classifying the type of metadata problem, etc. but I'm assuming that's beyond the scope of what is needed now.

### Administration
I'm also assuming that for the time being at least, we don't need to create an admin interface for managing incoming suggestions. If we need it, I don't mind mocking something up, but it can get pretty complex so I'll not do it unless necessary. As a general example, though, for Revs we have a dashboard for managing correction submissions (important there, given we've had 97,000 suggestions):

![revs-flagged-items-dashboard](https://user-images.githubusercontent.com/101482/29641500-b9662eaa-8818-11e7-9122-e2da7b2ae9b2.png)

For Revs, while you don't have to be logged in to provide a metadata suggestion, if you are logged in, you have the option to receive an email when your suggestion is resolved. This adds complexity to the submission form, since we need to both offer the email option to the user and give the administrator an associated form to record the resolution:

![revs-resolving-flag](https://user-images.githubusercontent.com/101482/29641534-e64c4bb6-8818-11e7-8a6d-dacc26a37367.png)

anarchivist:
Let's discuss on Thursday's standup?

mwerla:
[...]

PR #1152, opened 2021-01-27
Update some Arabic translations now that we've received them

jkeck:
## Contact Form (before)
contact-form-before

## Contact Form (after)
contact-form-after

---

## Collapse Button (before)
collapse-before

## Collapse Button (after)
collapse-after

---
## Record Feedback Form (before)
feedback-before

## Record Feedback Form (after)
feedback-after

### DIFF
--- a/app/views/record_feedback/new.html.erb
+++ b/app/views/record_feedback/new.html.erb
@@ -17,7 +17,7 @@
 
       <div class="modal-footer">
         <%= button_tag(t('.close'), class: 'btn btn-outline-dark', data: { 'dismiss': 'modal' }) if request.xhr? %>
-        <%= submit_tag t('.submit'), class: 'btn btn-primary' %>
+        <%= f.submit class: 'btn btn-primary' %>
       </div>
     <% end %>
   <% end %>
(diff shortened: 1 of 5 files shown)


jkeck:
@ggeisler One question that came up when reviewing some of these.  While I understand that we're still waiting on translations for the form labels and whatnot, could/should I be using some existing labels for the Submit/Close buttons on this form?  We use the word "Send" (which is translated) and may be more appropriate here, but I'm not sure we have an equivalent for "Close".  So a potential proposal would be to use the same translations as "Send" but I think we'll need to ask for a "Close" translation as part of the ask on #146

Screen Shot 2021-01-27 at 10 03 41 AM

ggeisler:
@jkeck I'm pretty sure you could re-use any previous translation of Submit or Close. I don't think there's anything about this context that would result in a different translation for those terms if they are used elsewhere in the app.

But if we don't have Submit, your suggestion for using "Send" sounds fine to me. I think that works just as well or better in this context than "Submit." 👍

cbeer (approved):

Issue #1136, opened 2021-01-20
Improve design of gallery view


jacobthill:
There is a lot of variation in the size of thumbnails provided from data providers, which results in a messy presentation of results in the gallery view.

![image](https://user-images.githubusercontent.com/562200/105093184-5f553e80-5ab3-11eb-83b8-2ba9e55fb094.png)

Can we add a consistently sized/styled container to create a cleaner presentation? See, for example, LoCs handling of different proportions of thumbnails:

![image](https://user-images.githubusercontent.com/562200/105093260-801d9400-5ab3-11eb-92b5-530f77702998.png)

ggeisler:
@mwerla @jacobthill I've updated the gallery view of the results page in my local dev environment and made some screenshots. The layout is very similar to the LOC example above. I made screenshots with three different background colors for the box that contains the thumbnail image and caption (item title plus metadata fields). Basically, I started with white, but am not sure if the contrast with the site background color is too much (maybe a bit distracting). So I tried two dark variations. One uses the same background color as the facets that are adjacent to the result items. The other is pretty similar but has some blue in it (it's a very light variation of the blue in the site masthead).

Here's an example:

<img width="1233" alt="FFF-p-5" src="https://user-images.githubusercontent.com/101482/105259232-a6d9da00-5b48-11eb-8e66-b38d268934fc.png">

Go here for more examples and to compare the three color variations: https://projects.invisionapp.com/boards/SZ41JIUJK7Q/

A couple of notes:

- I only modified CSS, so I think we could do it relatively easily, if we decide we want to
- As Marcin mentioned in #1084, I added Type, Date, Creator, and Identifier as metadata fields to display. We might not want all of those, but it's easy to turn on and off different fields so Jacob could experiment to get the right set you want to show.
- Also from #1084, I hid the metadata field label from the display. Note that if we do this it has to be for all fields (i.e., we can't show/hide the label for some fields and not others)

mwerla:
@ggeisler Thanks, the new design looks much clearer to me. Regarding colors, I like the darker variants, but I don't have a strong preference for which of the two darker ones we could go.

mwerla:
There are issues with how long titles are handled - see below. I propose to cut the titles after +/- 3 lines.

https://dev.dlmenetwork.org/library/browse/group/manuscripts/astronomy-manuscripts

![image](https://user-images.githubusercontent.com/562200/106441512-2fcb0c80-648b-11eb-800b-1604aab28e53.png)

[...]

PR #1190, opened 2021-02-02
Limit height of gallery view thumbnails

ggeisler:
This is a first-pass at addressing @mwerla's [concern about the height of thumbnail images](https://github.com/sul-dlss/dlme/issues/1136#issuecomment-771598801) in gallery view. We can adjust the `max-height` value later after we assess the tradeoffs with the value set in this PR.

The issue addressed is that when there is a very vertically-oriented image in the gallery view results, it creates a very tall thumbnail area and forces the other items in its row to stretch to a unusually tall height. I went with limiting the height of thumbnail images to `200px` in this PR, which is a compromise between limiting the height of images while not reducing the width too much.

### Before example
Indicated item creates very tall row with a lot of empty space in its neighbor items.

Screen Shot 2021-02-02 at 9 12 31 AM

### After example
Indicated item's height is limited, which also reduces its width, but not so much that it isn't still useful as a visual preview. Neighbor items are now more reasonable in height.

Screen Shot 2021-02-02 at 9 13 14 AM

### Additional example
@mwerla Note that to get equal heights of images we'd have to go really far in limiting image height. For instance, below the setting is `max-height: 100px;` but in my opinion this is too drastic, as it reduces the width of tall items so much that there isn't much value in the image preview. And because it reduces the width of many other no-so-tall images, it creates a lot of whitespace around the images and reduces the overall visual appeal of the page.

Screen Shot 2021-02-02 at 9 14 12 AM

### DIFF
--- a/app/assets/stylesheets/blacklight_gallery.scss
+++ b/app/assets/stylesheets/blacklight_gallery.scss
@@ -17,6 +17,7 @@
       border-radius: 0;
       display: block;
       margin: 0 auto;
+      max-height: 200px;
     }
 
     .caption {


mejackreed (approved):

mwerla:
@ggeisler Ok, I get your point. Let's keep it as you propose then.

PR #1227, opened 2021-02-05
Override styling of data provider HTML within metadata values

### LINKED ISSUE #1224: Question: Items with headings display large font on item cards, as opposed to "normal size" font that displays on other item cards
Is this font size display disparity acceptable?

https://dev.dlmenetwork.org/library/catalog?f%5Bcho_type_facet.en_ssim%5D%5B%5D=Sound 

Screen Shot 2021-02-04 at 1 16 57 PM

ggeisler:
Fixes #1224.

Resets styling of some HTML tags found in data provider metadata so results and item show pages don't have odd visual displays.

This PR only targets heading and `<strong>` elements; we can add other elements if and when we find other problems.

There is still often weird spacing (unnecessary blank lines) in metadata, as can be seen in the After screenshots below, but those are empty elements in the provided metadata and aren't as easily solved with CSS updates.

## Before

before-results

---

before-show

## After
after-results

---

after-show

### DIFF
--- a/app/assets/stylesheets/dlme.scss
+++ b/app/assets/stylesheets/dlme.scss
@@ -334,6 +334,18 @@ body {
   transform-origin: right bottom;
 }
 
+// Override styling of data provider HTML
+// within results and show page metadata fields
+.document-metadata {
+  h1, h2, h3, h4, h5, h6 {
+    font-size: $font-size-base;
+  }
+
+  strong {
+    font-weight: normal;
+  }
+}
+
 // Overrides for mobile viewports
 @media (max-width: 767px) {
   .navbar-default .navbar-form {


mejackreed:
One question we should ask, should we be just be stripping certain markup in description fields? Seems like we might be playing wackamole

ggeisler:
What's the alternative? Some markup could be useful in the context of metadata (italics, for example).

cbeer:
I'm all for stripping markup (or.. say, mapping h1-h6 to strong? i'm not sure how tractable that is..). I'd guess including headers from the metadata as-is does weird things for accessibility too.

ggeisler:
I'd prefer not to map h1-h6 to strong because in the records that prompted the ticket about this long strings are marked up with strong and that looks odd to me in these contexts, even without the sizing of the heading tags.

That said, I'm happy to leave the decision up to the developers. Whether you want to strip all HTML instead of the selective restyling in this PR, or some other solution. It seems like we should do something, though, since Jacob sees no problem with overriding data provider HTML and not doing anything results in some pretty ugly records.

If and when someone has another plan, I'm happy for you to close this PR in favor of that PR.

mejackreed:
Got it, by not striping headings we end up having heading level accessibility issues that we cannot address in any other way.

@jacobthill is it possible to sanitize HTML on the transform side?

jacobthill:
I should be able to remove the html at harvest or transform.

ggeisler:
Removing the HTML is a much better strategy, for sure. Hopefully it'll take care of the empty elements we currently get, which creates unnecessary blank lines.

ggeisler:
And @cbeer sorry I missed your point earlier about the a11y issues with just restyling the h1-h6. I think that's correct and a good reason to strip the HTML completely if we can.

PR #1237, opened 2021-02-18
Spin up an nginx thing to proxy the Rails app

mejackreed:


anarchivist:
@aaron-collier Do you remember any context about why this was done? Was this about injecting custom headers, or something else?

PR #1261, opened 2021-05-12
Use Google Tag Manager to manage analytics

### LINKED ISSUE #1220: Embed Google analytics with Google Tag Manager
From Marcin: 

> One more thing – I just realized, that it would be great to switch the way DLME is connected to Google Analytics.
> 
> Currently, the GA tracking code is embedded in the DLME code. What we find much more flexible and powerful, is to use Google Tag Manager instead. It allows injecting any kind of tracking on the page, without asking the tech team for any work. It also makes it much easier to track things like scrolling depth on a page etc.
> 
> If you don’t have experience with GTM, I can easily create a configuration for DLME and send you the tracking code that should be placed instead of the current GA tracking code. The GA data collection would be still going to the same GA account, so we wouldn’t loose the continuity of data.

> Paste this code as high in the `<head>` of the page as possible:
> 
> ```
> 
> 
> <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
> 
> new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
> 
> j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
> 
> 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
> 
> })(window,document,'script','dataLayer','GTM-T96MPMB');</script>
> 
> 
> ```
> 
> Additionally, paste this code immediately after the opening `<body>` tag:
> 
> ```
> 
> 
> <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-T96MPMB"
> 
> height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
> 
> 
> ```
> 
> Please let me know once that’s done in production, so I can verify if all works. I can provide similar code for dev environment if it’s easy to have different code snippets depending on the environment (there will be just a different GTM id).
> 
>

cbeer:
Fixes #1220

- [x] get GTM ids for all environments
- [x] decide if we want to make a better seam for analytics upstream
- [ ] decide if we want to support GTM upstream (e.g. in `spotlight/users/analytics.js`) or make a seam for it
- [x] find out where we can actually put the GTM tags; override the whole layout just to put them in the specified places is a little annoying / a maintainability risk.

### DIFF
--- a/app/views/layouts/spotlight/base.html.erb
+++ b/app/views/layouts/spotlight/base.html.erb
@@ -22,6 +22,7 @@
     <%= javascript_tag '$.fx.off = true;' if Rails.env.test? %>
   </head>
   <body class="<%= render_body_class %>">
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=<%= Settings.analytics.gtm_id %>" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
     <div id="skip-link">
       <%= link_to t('blacklight.skip_links.search_field'), '#search_field', class: 'element-invisible element-focusable rounded-bottom py-2 px-3', data: { turbolinks: 'false' } %>
       <%= link_to t('blacklight.skip_links.main_content'), '#main-container', class: 'element-invisible element-focusable rounded-bottom py-2 px-3', data: { turbolinks: 'false' } %>
--- a/app/views/layouts/spotlight/base.html.erb
+++ b/app/views/layouts/spotlight/base.html.erb
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<%= content_tag :html, class: 'no-js', **(try(:html_tag_attributes) || {}) do %>
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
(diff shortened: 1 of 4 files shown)


app/views/layouts/spotlight/base.html.erb
```diff
@@ -0,0 +1,49 @@
...
+    <%= javascript_include_tag "application" %>
+    <%= csrf_meta_tags %>
+    <%= content_for(:head) %>
+    <%= description %>
+    <%= twitter_card %>
+    <%= opengraph %>
+    <%= javascript_tag '$.fx.off = true;' if Rails.env.test? %>
+  </head>
+  <body class="<%= render_body_class %>">
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-T96MPMB" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
```
mjgiarlo, line 25: @cbeer What's this GTM ID?

mjgiarlo: @cbeer oh, nvm, that's the production ID. I don't think we'd want to hardcode that here, but maybe I'm mistaken.

camillevilla (approved):

PR #1285, opened 2021-06-17
Update application_profile.md

jacobthill:
## Why was this change made?

Adding new fields/values to the data model.

yes

### DIFF
--- a/docs/application_profile.md
+++ b/docs/application_profile.md
@@ -54,6 +54,7 @@ The DLME Prototype Application Profile a subset of the [Europeana Data Model](ht
 
 | Solr field name  | Property          | Expectation | Cardinality | Value type  | Definition | Notes                      |
 | ---------------- | ----------------- | ----------- | ----------- | ----------- | ---------- | -------------------------- |
+| cho_aat_material | `http://vocab.getty.edu/page/aat/300010357` | R | 0...n | literal | Getty Art & Architecture material facet. | |
 | cho_alternative  | `dct:alternative` |             | 0...n       | literal     | Alternative title for the resource. | |
 | cho_contributor  | `dc:contributor`  | R           | 0...n       | literal     | Contributor to the existance of the resource.  | |
 | cho_coverage     | `dc:coverage`     | R           | 0...n       | literal     | The spatial or temporal topic of the resource, the spatial applicability of the resource, or the jurisdiction under which the resource is relevant. | |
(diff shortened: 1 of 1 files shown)


docs/application_profile.md
```diff
@@ -54,14 +54,15 @@ The DLME Prototype Application Profile a subset of the [Europeana Data Model](ht
 
 | Solr field name  | Property          | Expectation | Cardinality | Value type  | Definition | Notes                      |
 | ---------------- | ----------------- | ----------- | ----------- | ----------- | ---------- | -------------------------- |
+| cho_aat_material | `http://vocab.getty.edu/page/aat/300010357` | R | 0...n | literal | Getty Art & Architecture material facet. | |
```
anarchivist, line 57: The property column should match to a specific RDF property; I can take a look and make a recommendation later.

docs/application_profile.md
```diff
@@ -70,6 +71,7 @@ The DLME Prototype Application Profile a subset of the [Europeana Data Model](ht
 | cho_is_part_of   | `dct:isPartOf`    |             | 0...n       | literal/ref | A related resource in which the described resource is physically or logically included. | |
 | cho_language     | `dc:language`     | M+          | 0...n       | literal     | Language for the resource. | Mandatory for text in Europeana; determine if BCP47 language tags are feasible |
 | cho_medium       | `dct:medium`      |             | 0...n       | literal     | The material or physical carrier of the resource. | |
+| cho_periodo_period | | R | 0...n | literal/ref | URI for the periodo period matching the cho_temporal string value | |
```
anarchivist, line 74: The property column should match to a specific RDF property; I can take a look and make a recommendation later.

anarchivist (changes requested):

docs/application_profile.md
jacobthill: is it `aat:300010357`?

anarchivist: No, that's for "Materials" as a specific facet within the AAT hierarchy. Is this intended to be a controlled value? It feels similar to `dct:medium` as used below (`cho_medium`) although I understand that we wouldn't necessarily want to duplicate use of that.

docs/application_profile.md
anarchivist: @jacobthill Is the intent to have a Periodo URI associated with the value in `cho_temporal`?

anarchivist: If that's the case, I think maybe `edm:hasMet` is the right property here?

docs/application_profile.md
jacobthill: yes it is a controlled value; it is our effort to normalize cho_medium. The values are from the AAT Materials facet.

docs/application_profile.md
jacobthill: Yes, that is the intent. If the string in the cho_temporal field matches a string in the periodo data, I fetch the periodo uri/s. `edm:hasMet` might be right though I'm not sure. Let me see if that makes sense to Wayne and Marcin.

PR #1301, opened 2021-08-05
Delete all resources

jmartin-sul:
## Why was this change made?

closes sul-dlss/dlme-transform#702

a description was added for the new rake task that makes this new functionality available.  can also add to the README, if desired.  i didn't see anything similar in the README, but this also seems like the app's first custom rake task?

### DIFF
--- a/app/jobs/delete_resources_job.rb
+++ b/app/jobs/delete_resources_job.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Load JSON resources from a url, persist them, and index them
+# Load JSON resources from a url, then remove them from the index and delete them
 class DeleteResourcesJob < ApplicationJob
   queue_as :default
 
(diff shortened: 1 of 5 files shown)


app/jobs/delete_resources_job.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Load JSON resources from a url, persist them, and index them
+# Load JSON resources from a url, then remove them from the index and delete them
```
jmartin-sul, line 3: this was just a fix to what i think was some old copypasta

app/jobs/delete_resources_job.rb
```diff
@@ -9,35 +9,14 @@ def perform(url, exhibit)
...
-    end
-
-    logger.info("#{resources.count} records were removed from #{url}.")
-  end
-
-  private
-
-  def remove_resource(item:, exhibit:)
-    resource = DlmeJson.find_by(url: item['id'], exhibit: exhibit)
-    return unless resource
```
jmartin-sul, line 24: pulled this into the loop in `#perform`

app/jobs/delete_resources_job.rb
```diff
@@ -9,35 +9,14 @@ def perform(url, exhibit)
...
 
-    # This approach will work for DLME where we only have one exhibit, but
-    # will need to be made more complicated for Spotlight in general because
-    # one resource could be part of more than one exhibit.
-    blacklight_solr.delete_by_query(
-      "{!terms f=#{Spotlight::Engine.config.resource_global_id_field}}#{resource.to_global_id}"
-    )
-    blacklight_solr.commit
-    resource.destroy
-  end
```
jmartin-sul, line 34: refactored the rest of the private method functionality into a service class, to re-use in the new rake task

app/services/resource_remover.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    removal_count
+  end
+
+  def self.remove_resource(resource:)
+    new.remove_resource(resource: resource)
+  end
+
+  # @param [Spotlight::Resource] resource to be un-indexed and destroyed
+  def remove_resource(resource:)
```
jmartin-sul, line 26: i didn't add tests for this method to the new service class' specific tests, since the existing `DeleteResourcesJob` tests cover it, but can revisit that if others think that'd be a good idea.

jmartin-sul: the new `remove_all_resources` test implicitly exercises this too

aaron-collier (approved):

Issue #1303, opened 2021-08-10
Add Delete action to Bulk actions menu


ggeisler:
@jacobthill @aaron-collier Based on earlier Slack discussion, if the team decides to enable a delete items action using the Bulk actions menu, here's a suggested design:

## New Bulk actions menu item

<img width="380" alt="Screen Shot 2021-08-09 at 5 32 42 PM" src="https://user-images.githubusercontent.com/101482/128791006-e5020151-d58f-41d6-a395-a575db761c0c.png">

## Delete items modal
Put the Delete button in the main modal area to minimize accidental activation that might happen if we put it in the modal footer as we do the main action button of the other modals. It makes things a bit inconsistent but this action has larger negative ramifications than the other actions:

<img width="524" alt="Screen Shot 2021-08-09 at 5 50 53 PM" src="https://user-images.githubusercontent.com/101482/128792047-e47ebb36-77c5-416a-86e8-58139480012d.png">

---

- Colors/fonts in the modal should be the same as the current modals in DLME (I used an Exhibits mockup for this so colors/font don't match)

- The red area in the modal uses the [Bootstrap alert styling](https://getbootstrap.com/docs/5.0/components/alerts/)

## Confirmation modal
Below is the confirmation modal from a Stanford exhibit. I think the only thing that needs to be updated here is changing the word "updated" to "deleted":

<img width="531" alt="Screen Shot 2021-08-09 at 5 43 07 PM" src="https://user-images.githubusercontent.com/101482/128791566-a8eb6d68-dd6f-4d5c-8780-8e86ec3f1c21.png">

PR #1310, opened 2021-08-13
delete items using search results

jmartin-sul:
## Why was this change made?

closes sul-dlss/dlme-transform#703
see also https://github.com/sul-dlss/dlme/issues/1303

Screen Shot 2021-08-19 at 3 54 58 PM
Screen Shot 2021-08-19 at 3 55 14 PM

N/A?  but happy to update docs if it turns out i should.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -8,6 +8,8 @@ class CatalogController < ApplicationController
   extend MultilingualLocaleAwareField
 
   before_action do
+    blacklight_config.bulk_actions&.delete_resources!
+
     next unless request.format.json?
 
     blacklight_config.add_show_field :__raw_resource_json_ss, helper_method: :ir_for_output
--- a/app/views/spotlight/catalog/_delete_resources.html.erb
+++ b/app/views/spotlight/catalog/_delete_resources.html.erb
@@ -1,0 +1,40 @@
+<div class="modal fade" id="delete-resources-modal" tabindex="-1" role="dialog" aria-labelledby="save-modal-label" aria-hidden="true">
+  <div class="modal-dialog">
+  <%= bootstrap_form_for([:delete_resources, current_exhibit], url: delete_resources_exhibit_dlme_bulk_actions_path(current_exhibit), method: 'post') do |f| %>
+    <div class="modal-content">
+      <div class="modal-header">
+        <h4 class="modal-title" id="save-modal-label"><%= t(:'spotlight.bulk_actions.delete_resources.heading') %></h4>
+        <button type="button" class="blacklight-modal-close close" data-dismiss="modal" aria-label="<%= t('blacklight.modal.close') %>">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <%= render Blacklight::HiddenSearchStateComponent.new(params: search_state.params_for_search) %>
+        <p>
+          <%= t(:'spotlight.bulk_actions.delete_resources.description', count: @response.total) %>
+        </p>
+
+        <div class="alert alert-danger container" role="alert">
+          <div class="row">
+            <div class="col-1">
+              <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-exclamation-triangle" viewBox="0 0 16 16">
+                <path d="M7.938 2.016A.13.13 0 0 1 8.002 2a.13.13 0 0 1 .063.016.146.146 0 0 1 .054.057l6.857 11.667c.036.06.035.124.002.183a.163.163 0 0 1-.054.06.116.116 0 0 1-.066.017H1.146a.115.115 0 0 1-.066-.017.163.163 0 0 1-.054-.06.176.176 0 0 1 .002-.183L7.884 2.073a.147.147 0 0 1 .054-.057zm1.044-.45a1.13 1.13 0 0 0-1.96 0L.165 13.233c-.457.778.091 1.767.98 1.767h13.713c.889 0 1.438-.99.98-1.767L8.982 1.566z"/>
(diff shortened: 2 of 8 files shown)


spec/features/bulk_resource_delete_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    end
+  end
+
+  context 'with an ordinary user' do
+    it 'does not offer bulk delete' do
+      visit root_path
+      click_button 'Search'
+      expect(page).not_to have_content 'Bulk actions'
+      expect(page).not_to have_content 'Delete items'
+    end
```
jmartin-sul, line 124: i don't know if it's worth the time, since the behavior is inherited from spotlight, and seems to work as expected here... but i had also considered adding a request spec that shows that authorization checks are enforced when hitting the new controller method for bulk delete (so as to prevent savvy but maliscious users from constructing a delete request despite the lack of button in the UI).

spec/features/bulk_resource_delete_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+      'traject_context_source' => 'not_in_the_original_search_results' }
+  end
+  let(:metadata_3) do
+    { 'traject_context_command_line.filename' => fixture_file_path_3,
+      'traject_context_source' => 'not_in_the_original_search_results' }
+  end
+  let(:metadata_4) do
+    { 'traject_context_command_line.filename' => fixture_file_path_4,
+      'traject_context_source' => 'matching_resource' }
+  end
```
jmartin-sul, line 35: sorry, this is rather verbose and repetitive.  happy to revisit and find some way to make more concise if people want.  but i did want the tests to show that exactly the subset of things in the search results gets deleted.

config/routes.rb
```diff
@@ -35,6 +35,12 @@
       resource :s3_delete, controller: :s3_delete, only: [:new, :create]
       resource :statistics, only: :show
 
+      resource :dlme_bulk_actions, only: [] do
+        member do
+          post :delete_resources
+        end
+      end
```
jmartin-sul, line 42: sorta copypasta'd from the spotlight `routes.rb`

app/views/spotlight/catalog/_delete_resources.html.erb
```diff
@@ -0,0 +1,40 @@
...
+          </div>
+        </div>
+      </div>
+      <div class="modal-footer d-flex flex-row-reverse justify-content-start">
+        <button type="button" class="btn btn-primary" data-dismiss="modal"><%= t :cancel %></button>
+      </div>
+    </div>
+    <% end %>
+  </div>
+</div>
```
jmartin-sul, line 40: all copypasta'd from a spotlight bulk action

app/controllers/catalog_controller.rb
```diff
@@ -8,6 +8,8 @@ class CatalogController < ApplicationController
   extend MultilingualLocaleAwareField
 
   before_action do
+    blacklight_config.bulk_actions&.delete_resources!
```
jmartin-sul, line 11: overridden here because a) `CatalogController` seems to be the only thing where `blacklight_config.bulk_actions` exists, and b) because the catalog view page is the only place it shows up anyway (right?).

app/views/spotlight/catalog/_delete_resources.html.erb
```diff
@@ -0,0 +1,40 @@
...
+          <%= t(:'spotlight.bulk_actions.delete_resources.description', count: @response.total) %>
+        </p>
+
+        <div class="alert alert-danger container" role="alert">
+          <div class="row">
+            <div class="col-1">
+              <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="currentColor" class="bi bi-exclamation-triangle" viewBox="0 0 16 16">
+                <path d="M7.938 2.016A.13.13 0 0 1 8.002 2a.13.13 0 0 1 .063.016.146.146 0 0 1 .054.057l6.857 11.667c.036.06.035.124.002.183a.163.163 0 0 1-.054.06.116.116 0 0 1-.066.017H1.146a.115.115 0 0 1-.066-.017.163.163 0 0 1-.054-.06.176.176 0 0 1 .002-.183L7.884 2.073a.147.147 0 0 1 .054-.057zm1.044-.45a1.13 1.13 0 0 0-1.96 0L.165 13.233c-.457.778.091 1.767.98 1.767h13.713c.889 0 1.438-.99.98-1.767L8.982 1.566z"/>
+                <path d="M7.002 12a1 1 0 1 1 2 0 1 1 0 0 1-2 0zM7.1 5.995a.905.905 0 1 1 1.8 0l-.35 3.507a.552.552 0 0 1-1.1 0L7.1 5.995z"/>
+              </svg>
```
jmartin-sul, line 23: this svg is the warning triangle in the delete confirmation dialog.  i'd initially tried to render it using just bootstrap classes, but that didn't work for me, so i used their suggested svg.  also happy to revisit this if there's a better way (or if i was just doing something wrong the other way i tried).

jmartin-sul (commented):
some inline explanation

config/locales/ar.yml
```diff
@@ -65,6 +65,9 @@ ar:
     new:
       add_item: احذف العناصر
   spotlight:
+    bulk_actions:
+      delete_resources:
+        heading: حذف الموارد
```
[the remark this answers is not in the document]
jmartin-sul: i filed a ticket: https://github.com/sul-dlss/dlme/issues/1313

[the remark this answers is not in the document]
jacobthill: Thanks, @jmartin-sul. This shouldn't hold up merging. QNL will need to get involved and it will be slow. These are also fields that should only be visible to those with admin privileges and currently we have no need for these fields since we don;t have Arabic speaking curators.

aaron-collier:
@jmartin-sul Jumping into a full review, but manual testing this is working :+1: exactly how I imagined. I think this will be a HUGE improvement for @jacobthill

aaron-collier (approved):

jmartin-sul:
re-based on latest `main`

PR #1316, opened 2021-08-24
Add a simple rake task for purging transform results

aaron-collier:
## Why was this change made?

This is a small rake task to clear out existing transform results. This will be run as a one off ECS task when needed so there is no need to expose this in the UI.


ndushay (approved):
LGTM????

Issue #1177, opened 2021-02-01
Spacing of text on item cards in gallery view results in partially obscured text


caaster:
Please see the example below:
![Screen Shot 2021-02-01 at 9 50 19 AM](https://user-images.githubusercontent.com/3210225/106497466-ed95d100-6472-11eb-8f31-0130c6ebde0f.png)

jkeck:
The strategy that has been implemented for truncating this content at a particular number of lines is to use line-height & the height of the element to calculate the number of lines present in an element.

The reason this is happening is because the particular field in question has arbitrary HTML contained within (which can have different font sizes/padding/etc). There may be an element that introduces some sort of sizing that may throw off the line calculation (in the example above it's the bottom margin on the `<p>` tags).  

One possible improvement might be to add a gradient effect to the bottom of a truncated element so that it "fades" into the element.  This would at least obscure more of the text (and might be less visually jarring).

Another potential avenue to look into would be to look into the element's direct children and use the bottom of any element that intersects the goal height to actually set the height.  This seems like it would be quite a bit more complicated to implement.

ggeisler:
Noting that a minor related issue I noticed when watching some Hotjar sessions: occasionally the "more" toggle is displayed when there actually is no more text to display (I guess because the text is just the right length to trigger the need for the toggle, but because of the challenge Jessie described above, the text is actually completely displayed).

This isn't a major UX issue, but I did see a couple of users click on it and it revealed nothing additional, so they probably were at least mildly annoyed at having selected something that had no effect.

ggeisler:
Given:

- My comment above noting that sometimes the calculation of whether there is more text to display is off slightly (presumably because of the presence of arbitrary HTML in the text, as Jessie suggested is the source of the problem), resulting in a "More" button with nothing more to display, and
- The challenge of finding an obvious way to mask partial lines of text that might or might not be immediately above the button (I've tried a bunch of CSS tricks to deal with this, but none seem feasible)

I wonder if the best solution is simply to strip out the arbitrary text from the HTML that is causing the problem? I guess I've lost track of why we think we need to preserve HTML text in metadata that comes from data providers, but if we  strip out at least selected HTML tags (i.e., only tags that affect font sizing) from the search result metadata before we display it, wouldn't that ensure that the currently imperfect calculations related to the "More" button would be more accurate?

ggeisler:
[...]

PR #1413, opened 2022-02-03
Remove margin-bottom from paragraphs in thumbnail description

marlo-longley:
## Why was this change made?

This change addresses #1177 by removing the `margin-bottom` on paragraph tags in the thumbnail view.
There were a couple ways to address this, but after making some height calculations in Javascript and realizing this same markup is shared between different views, I feel this is the simplest solution. The paragraph margins are left as is in the full item view for readability. 

### Before
Screen Shot 2022-02-03 at 15 15 23

### After
Screen Shot 2022-02-03 at 15 14 50

### Full Item View
Screen Shot 2022-02-03 at 15 15 07

N/A


ggeisler (approved):
Looks like that works. Thanks for the screenshots @marlo-longley.

PR #1444, opened 2022-02-09
Configure additional facet fields.

### LINKED ISSUE #1326: Expose more fields as facets in the DLME web app
As a DLME curator I would like to browse DLME records by additional fields, some which may be useful to end users and others that have curatorial value but users will not be interested in. 

Expose the following fields as facets in the [curator view](https://dev.dlmenetwork.org/library/search_configuration/edit):
- cho_coverage
- agg_data_provider_collection <strike>dlme_colection</strike>
- agg_data_provider_collection_id <strike>dlme_colection_id</strike>
- cho_subject
- agg_edm_rights
- agg_dc_rights
- wr_dc_rights
- wr_edm_rights

corylown:
Closes #1326


cbeer:
I'm not sure how many of these make sense to show to end users (and/or, good luck to us giving them labels that a user can understand without intimate knowledge of the IR). I assume that plan is to ship this and then refine the list later?

cbeer:
👍 sounds good. I guess with this, Jacob can show or hide whatever from the curator interface 🤷‍♂️

PR #1455, opened 2022-02-10
JSON import tab UI

### LINKED ISSUE #1400: Display, but disable, "DLME JSON" tab when not applicable
I'm not certain about the technical feasibility of this, nor whether it is a priority (since it only affects admin users), but the team did note a confusing aspect of the admin UI yesterday that ideally we'd clarify to prevent similar confusion in the future.

On the Curation > Add items > From external resource tab, Docker users see two tabs:

Screen Shot 2022-02-01 at 15 03 50

while non-Docker users only see the second tab, "DLME S3 Fetch":

Screen Shot 2022-02-01 at 4 04 50 PM

---

Ideally, non-Docker users should see the same tabs, with the "DLME JSON" tab disabled rather than hidden (if it makes things easier, I don't see any problem switching the order of the two tabs and making the "DLME Fetch" tab the first and default selection, since that's the one that should always be enabled):

Screen Shot 2022-02-02 at 9 51 55 AM

---

It also appears that the "DLME S3 Fetch" option is actually not specific to **S3** URLs, so in the mockup above I've also suggested we relabel that tab to "DLME Fetch."

thatbudakguy:
- Change label for DLME JSON fetch tab
- Restructure external resource form for more control over HTML
- Hide external resource form tabs when there's only one
- Update strings for external resource importing

closes #1400


corylown (approved):
This all makes sense to me and it looks like it does what it's supposed to do :-)

PR #1463, opened 2022-02-14
Test built docker image in CI

thatbudakguy:
- Specify solr and postgres versions for local dev
- Reformat docker-compose file
- Persist solr data using named volume
- Use entrypoint script to run rails server for app
- Move args/env to beginning of docker builds
- Use envvar to tell bundler what dependencies to omit
- Use ENTRYPOINT command for worker build
- Add more docker info to README
- Remove version check in entrypoint since it's not set
- Update redis version in docker-compose.yml
- Take advantage of Docker build stage targets
- Update image names for compose build
- Remove some unneeded quotes
- Remove compose entrypoint for webapp
- Update the entrypoint script to use rails bin/setup
- Remove old rails 5 bin/update
- Reformat envvars syntax in compose file
- Update docker setup instructions in README
- Use build target stages in CircleCI
- Create envvar that makes Capybara point at Docker
- Create compose stack for testing in CircleCI
- Add docker stack integration testing to CircleCI


thatbudakguy:
closing since we're moving on-prem and I never got this to work quite right.

PR #1485, opened 2022-02-22
Pad digits in the sortable title so they sort 'naturally'

### LINKED ISSUE #1192: Improve sort by title behavior
A summary of our current strategy for sorting by title (from @cbeer): "I'm not sure how we came up with it, but our current algorithm to pick a sortable title is to go with the first one from the IR, and in this case it happens to be the Arabic version. At a guess, I'd say we just picked the first, regardless of language, to keep a consistent sort order across locales (because it would probably be annoying to go... "ok, i want to see this page in Arabic now" and get a totally different result set?). I don't know if we can be smarter, or if the data can be remapped to include all the information in both the English + Arabic version of the title?" 

The current strategy (described above) seems reasonable, however, it would be ideal to parse strings for numbers and, if found, sort the numbers numerically. The most pressing use case is for Serials (e.g. [these](https://dlmenetwork.org/library/browse/group/serials/nashriyah-digital-iranian-history-u-manchester)).

cbeer:
Fixes #1192

### DIFF
--- a/app/resource_builders/dlme_json_resource_builder.rb
+++ b/app/resource_builders/dlme_json_resource_builder.rb
@@ -46,14 +46,18 @@ class DlmeJsonResourceBuilder
 
         source[key].each do |language_code, values|
           sink["#{key}.#{language_code}_tsim"] = values
-          sink["sortable_#{key}.#{language_code}_ssi"] = values.first
+
+          value = values.first
+          value = apply_natural_sort_munging(value) if key.starts_with?('cho_title')
+
+          sink["sortable_#{key}.#{language_code}_ssi"] = value
         end
       else
         sink["#{key}_tsim"] = source[key]
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/resource_builders/dlme_json_resource_builder.rb
```diff
@@ -94,4 +98,11 @@ def transform_to_untokenized_solr_fields(source = {}, sink: {}, prefix: '')
   def field_suffix_for(field_name)
     FIELD_SUFFIXES.fetch(field_name, DEFAULT_FIELD_SUFFIX)
   end
+
+  def apply_natural_sort_munging(string)
+    return unless string
+
+    # pad any digits we run into so they'll sort "naturally" (1, 2, ... 8, 9, 10)
+    string.gsub(/\d+/) { |digits| digits.rjust(10, '0') }
```
cbeer, line 106: 10 digits ought to be enough for everybody, right?

corylown (approved):

PR #1493, opened 2022-02-23
Set Solr min match parameter to 1 on boolean search

### LINKED ISSUE #1352: Boolean OR not working in DLME search
OR is working like AND as far as I can tell. Ensure that OR works as expected.

marlo-longley:
Fixes #1352 

@corylown

- adds `min_match_for_boolean` method to SearchBuilder class 
- adds new fixture json dat for testing
- tests search queries containing OR, AND, and NOT


app/models/search_builder.rb
```diff
@@ -12,4 +15,16 @@ class SearchBuilder < Blacklight::SearchBuilder
...
   #   end
+
+  self.default_processor_chain += [:min_match_for_boolean]
+
+  # Sets min match to 1 based on presence of boolean operators, overriding default set in Solr
+  def min_match_for_boolean(solr_parameters)
+    return unless blacklight_params &&
+                  blacklight_params[:q].present? &&
+                  blacklight_params[:q].respond_to?(:match) &&
+                  blacklight_params[:q].match(/\s(AND|OR|NOT)\s/)
```
corylown, line 26: Didn't catch this earlier, but we could use `match?` here (and the line above) instead of `match` because we just want a true/false response and don't actually care about the matched values.

cbeer: Maybe something like:

```suggestion
    return unless search_state.query_param&.match?(/\s(AND|OR|NOT)\s/)
```

app/models/search_builder.rb
```diff
@@ -12,4 +15,16 @@ class SearchBuilder < Blacklight::SearchBuilder
   #   def add_custom_data_to_query(solr_parameters)
   #     solr_parameters[:custom] = blacklight_params[:user_value]
   #   end
+
+  self.default_processor_chain += [:min_match_for_boolean]
+
+  # Sets min match to 1 based on presence of boolean operators, overriding default set in Solr
```
cbeer, line 21:
```suggestion
  # Sets `mm` (minimum should match) to 1 based on presence of boolean operators, overriding default set in Solr.
  # The default value in the request handler (`6<-1 6<90%`) allows for some slop in full-text queries,
  # but leads to unexpected behavior when a user expects to use boolean operators for more advanced
  # search queries. Setting `mm=1` means only 1 clause needs to match, but the lucene query parser (in 
  # edismax) will precompose the query strings so everything works out.
```

Is that a fair summary of why mm=1?

corylown: yes, I think 0% would work too, but 1 makes more sense to me for some reason.

app/models/search_builder.rb
corylown: applied suggestion

thatbudakguy (approved):
this is great; i really like the way the tests/fixtures are written: minimal enough to understand quickly but substantial enough to test the functionality.

Issue #1479, opened 2022-02-18
Configure the date sort options in the DLME app


corylown:
- Configure ability to sort results by date in ascending order: `Sort by Date (old to new)`. Ascending order sort uses the minimum date value derived from cho_date_range_norm_isim for sorting. Field is likely to be `cho_date_norm_min_isi`
- Configure ability to sort results by date in descending order: `Sort by Date (new to old)`. Descending order uses the maximum date value derived from cho_date_range_norm_isim for sorting. Field is likely to be `cho_date_norm_max_isi`

Blocked by:

- #1477 

Relates to:

- #1470 
- #1476 
- #1478

Partially addresses:

- #1348

corylown:
Can be QA'd if date sort is configured via admin/curator UI. Sort missing last is not yet implemented, but the work necessary for this feature is captured by another issue #1476

PR #1502, opened 2022-02-25
Configure date sort; add contextual alert.

### LINKED ISSUE #1470: Provide context for users on "sort by" functionality
As a DLME user, I sometimes am confused by the sort behavior. Ideally the UI would provide enough context for me to interpret the sort results.

In DLME the sort by functionality relies on data quality/completeness. We are planning on adding sort by date to the existing 'relevance', 'title' and 'creator' options via https://github.com/sul-dlss/dlme/issues/1348. This sort behavior will likely be confusing to users. Is there a way to clarify in the UI:

- [ ] That some records may be missing values and these records will always be at the end
- [ ] That dates are the earliest_dates provided by the data provider
- [ ] Ideally we can fix the way titles sort when numbers are part of the title https://github.com/sul-dlss/dlme/issues/1192 but this is related so linking it here

Screen Shot 2022-02-16 at 2 11 55 PM

corylown:
Closes #1470 and #1479

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -265,8 +265,10 @@ class CatalogController < ApplicationController
     config.add_sort_field 'relevance', sort: 'score desc, sortable_cho_title_ssi asc', label: 'Relevance'
     config.add_sort_field 'title', sort: 'sortable_cho_title_ssi asc, sortable_cho_creator_ssi asc', label: 'Title'
     config.add_sort_field 'creator', sort: 'sortable_cho_creator_ssi asc', label: 'Creator'
-    config.add_sort_field 'date_old_to_new', sort: 'cho_date_norm_min_isi asc', label: 'Date (old to new)'
-    config.add_sort_field 'date_new_to_old', sort: 'cho_date_norm_max_isi desc', label: 'Date (new to old)'
+    config.add_sort_field 'date_old_to_new', sort: 'cho_date_norm_min_isi asc, sortable_cho_title_ssi asc',
+                                             label: 'Date (old to new)'
+    config.add_sort_field 'date_new_to_old', sort: 'cho_date_norm_max_isi desc, sortable_cho_title_ssi asc',
+                                             label: 'Date (new to old)'
 
     config.add_field_configuration_to_solr_request!
   end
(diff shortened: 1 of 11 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -265,6 +265,8 @@ class CatalogController < ApplicationController
     config.add_sort_field 'relevance', sort: 'score desc, sortable_cho_title_ssi asc', label: 'Relevance'
     config.add_sort_field 'title', sort: 'sortable_cho_title_ssi asc, sortable_cho_creator_ssi asc', label: 'Title'
     config.add_sort_field 'creator', sort: 'sortable_cho_creator_ssi asc', label: 'Creator'
+    config.add_sort_field 'date_old_to_new', sort: 'cho_date_norm_min_isi asc', label: 'Date (old to new)'
```
cbeer, line 268:
```suggestion
    config.add_sort_field 'date_old_to_new', sort: 'cho_date_norm_min_isi asc, sortable_cho_title_ssi asc', label: 'Date (old to new)'
```

Just a thought.

app/controllers/session_displays_controller.rb
```diff
@@ -4,9 +4,10 @@
 # A controller for disabling session information
 class SessionDisplaysController < ApplicationController
   ##
-  # Update the session, to disable_search_context
+  # Update the session, to disable_search_context or disable_date_sort_context
   def update
-    session[:disable_search_context] = true
+    session[:disable_search_context] = true if params[:context] == 'result_info'
```
cbeer, line 9: Is it worth trying to align the session key with the `params[:context]` value?

app/views/shared/_contextual_result_info.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="alert alert-info alert-dismissible fade show" role="alert">
+<div class="result_info_message alert alert-info alert-dismissible fade show" role="alert">
```
cbeer, line 1: Is the `result_info_message` class used, or is it just there for our benefit? 

```suggestion
<div class="result-info-message alert alert-info alert-dismissible fade show" role="alert">
```

I think we prefer dashes in class names?

corylown: I added the class here for parity between the two alerts and I wanted the class on the date sort alert to select for testing 🤷‍♂️ . I'll switch to dashes

app/controllers/session_displays_controller.rb
corylown: makes sense to me

app/controllers/catalog_controller.rb
corylown: probably a good thought

cbeer: (and... why we didn't think that thought when we added `creator`... 🤷‍♂️ )

PR #1503, opened 2022-02-25
Document supported ruby version

thatbudakguy:
- Add ruby version badge to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 [![CircleCI](https://circleci.com/gh/sul-dlss/dlme.svg?style=svg)](https://circleci.com/gh/sul-dlss/dlme)
 [![Coverage Status](https://coveralls.io/repos/github/sul-dlss/dlme/badge.svg)](https://coveralls.io/github/sul-dlss/dlme)
+![Ruby Version](https://img.shields.io/badge/ruby-3.0-red)
 
 # Digital Library of the Middle East
 


marlo-longley:
@thatbudakguy thanks! This looks simple and helpful to me..

marlo-longley (approved):

README.md
```diff
@@ -1,5 +1,6 @@
 [![CircleCI](https://circleci.com/gh/sul-dlss/dlme.svg?style=svg)](https://circleci.com/gh/sul-dlss/dlme)
 [![Coverage Status](https://coveralls.io/repos/github/sul-dlss/dlme/badge.svg)](https://coveralls.io/github/sul-dlss/dlme)
+![Supported Ruby Version](https://img.shields.io/badge/dynamic/yaml?color=red&label=ruby&query=%24.parameters.ruby_version.default&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsul-dlss%2Fdlme%2Fdocument-ruby-version%2F.circleci%2Fconfig.yml)
```
cbeer, line 3:
```suggestion
![Supported Ruby Version](https://img.shields.io/badge/dynamic/yaml?color=red&label=ruby&query=%24.parameters.ruby_version.default&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsul-dlss%2Fdlme%2Fmain%2F.circleci%2Fconfig.yml)
```

Presumably we'd want to point this at `main`, not this branch?

cbeer:
-0; this seems like a complicated way to get "we're using ruby 2.7" into the README. 

I'm not sure what problem we're trying to solve, but I'd guess it's.. a desire was to make sure the versions stay in sync? .

README.md
thatbudakguy: oooh good catch, thx

thatbudakguy:
yup, just a desire to make the README match the version we actually test against/use. i also like it when major versions of things are at the top of the CI config file rather than buried, so in this case doing that made it easier to also the former thing

thatbudakguy:
i guess this doesn't solve keeping the version in the dockerfile in sync though, so maybe it's less effective than it could be. i'm fine with just making a regular badge in the README if you would prefer that @cbeer

thatbudakguy:
changed it to a regular badge for ruby v3.0.

PR #1504, opened 2022-02-25
Add Mirador Image Tools plugin

### LINKED ISSUE #1350: Add rotate plugin to Mirador viewer
Several users requested the ability to rotate images displayed in Mirador. This plugin exists but is not implemented in DLME.

marlo-longley:
Fixes #1350 

- Adds plugin to package.json
- Enables plugin in config
- Checks for plugin existence in spec

### DIFF
--- a/app/javascript/packs/mirador.js
+++ b/app/javascript/packs/mirador.js
@@ -21,10 +21,12 @@ import 'url-polyfill/url-polyfill';
 import 'unfetch/polyfill';
 
 import Mirador from 'mirador/dist/es/src/index.js';
+import miradorImageToolsPlugin from 'mirador-image-tools/es/plugins/miradorImageToolsPlugin.js';
+
 const manifestUrl = document.querySelector('#m3').dataset.iiifManifest;
 const htmlAttributes = document.querySelector('html').attributes;
 
-Mirador.viewer({
+const config = {
   id: 'm3',
   language: htmlAttributes.lang.nodeValue,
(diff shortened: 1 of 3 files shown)


app/javascript/packs/mirador.js
```diff
@@ -52,7 +54,15 @@ Mirador.viewer({
   workspaceControlPanel: {
     enabled: false,
   },
-  windows: [
-    {loadedManifest: manifestUrl}
-  ]
-})
+  windows: [{
+    imageToolsEnabled: true,
+    imageToolsOpen: true,
```
cbeer, line 59: @ggeisler : should we default to showing the whole toolbar or just the toggle? I'm not sure what the intended use for these tools is..

marlo-longley: Screen Shot 2022-02-25 at 12 00 31

Current view for reference with "open" as default

marlo-longley: Vs. closed default
Screen Shot 2022-02-25 at 12 01 20
t

ggeisler: I suggest closed by default. I think the goal for DLME is to enable users who might have a use case for one or more of the tools to have access to them, but there are likely far more DLME users who don't have image-tool-related use cases. So the more unobtrusive closed version seems appropriate as the default.

thatbudakguy (approved):

Issue #1109, opened 2021-01-13
Make better use of horizontal space


jacobthill:
On wider screens there is a lot of empty horizontal space
<img width="625" alt="Screen Shot 2021-01-13 at 8 40 36 AM" src="https://user-images.githubusercontent.com/37662787/104459697-1402d380-557b-11eb-8580-6afde744d756.png">

ggeisler:
I believe we have two main options here.

### Option 1 - Migrate to Bootstrap 5
This will give us BS5's [additional `xxl` breakpoint](https://getbootstrap.com/docs/5.0/layout/grid/#grid-options) of `1400px`. It would require considerable work than Option 2 below, and realistically should be done in Spotlight, which might also require some work in Stanford Exhibits to ensure everything looks good there after the upgrade. But this option also benefits the entire Spotlight community (including Stanford Exhibits) rather than just DLME.

### Option 2 - Add a new, wider breakpoint to DLME
This is likely a much simpler option, although it only benefits DLME. We would add a new breakpoint to DLME that will provide users who have wide viewports with more (or sometimes, larger) content horizontally compared to the current maximum viewport in DLME. If we pursue this option, I suggest we use `1400px` as the new breakpoint, to match the widest breakpoint added in BS5. That might make any customization we do invisible to existing DLME users when we do eventually migrate to BS5. I’m also not sure there is much value to going any wider with the new breakpoint; a wider value will create very long text lines on the home and about pages, and might require much more customization work if we wanted to add additional columns for tiles on the home, browse landing page, and search results gallery view pages.

### Customization details
Whichever option we go with, much of the DLME site should just work with the wider content area. Assuming that with whatever option we use the new breakpoint is `1400px`, that gives us a new `max-width` container of `1320px`, up from our current max of `1140px` (which is what the current `1200px` `xl` breakpoint gives us). Page elements will have more width to utilize, but usually that won't require any specific updates (e.g., paragraph text will just flow wider before breaking to a new line, but we don't have to do anything for that to happen). As noted below, there might be a few elements that need specific adjusting (e.g., through a media query).  Here's a screenshot of the homepage with a `1400px` breakpoint and the resulting `1320px` content area:

<img width="1904" alt="Screen Shot 2021-12-15 at 4 28 39 PM" src="https://user-images.githubusercontent.com/101482/146280967-56362af5-4576-4f2e-9db7-f20a45819805.png">


[...]

PR #1507, opened 2022-02-25
Add xxl breakpoint at 1400px

marlo-longley:
Minimally addresses #1109, not the entire issue - this PR simply adds a breakpoint at 1400px and lets the content fill out that space without any further refinements. It was suggested that we push this up to UAT for visual review, and if it's not approved, roll back this PR and wait to address this issue until a BS5 upgrade.

- 1400px (with 1320 max-width on containers) aligns with the xxl breakpoint in [BS5](https://getbootstrap.com/docs/5.0/layout/breakpoints/)
- by putting this edit into its own file, we can simply delete it when we upgrade to BS5

@ggeisler


cbeer:
Kapture 2022-02-28 at 07 47 32

PR #1510, opened 2022-02-28
Use agg_data_provider_collection_id_ssim for collection count.

### LINKED ISSUE #1327: Update how collections are counted in DLME
The current process for counting collections in DLME uses the file paths in dlme metadata which are passed to the `agg_data_provider_collection` field, but it isn't always accurate. In some cases (e.g. Sakip Sabanci) the collection is partially available from different sources and is spread over two folders when it is actually a single collection. Other collections have similar problems (e.g. QNL, AUC). The way the values are mapped has been updated and should be accurate. However, all values will be fully available in English and Arabic. It seems like the counting is based on all unique values in the `agg_data_provider_collection` field and it should be based on all unique values in the `agg_data_provider_collection` field in the English language. In other words, it should not count the collection name in English and its translation in Arabic as two separate collections. So either we need to parse the languages before counting or we could update the counter to use the `agg_data_provider_collection_id` field.

At the moment its not clear to me where collection counts are coming from. For example, looking at American University in Cairo on the [contributors page](https://dev.dlmenetwork.org/library/statistics) I see 23 collections, but if I click on the link to the data provider, there is no data in the `Collection ID` or the `Collection` facet. Maybe this is a caching issue?

corylown:
Closes #1327

### DIFF
--- a/app/models/statistics_dashboard.rb
+++ b/app/models/statistics_dashboard.rb
@@ -14,7 +14,7 @@ class StatisticsDashboard
       'agg_provider_country.en_ssim',
       'agg_provider_data_country.ar-Arab_ssim',
       'agg_provider_data_country.en_ssim',
-      'agg_data_provider_collection_ssim',
+      'agg_data_provider_collection_id_ssim',
       'cho_type_facet.en_ssim',
       'cho_type_facet.ar-Arab_ssim',
       'cho_language.en_ssim',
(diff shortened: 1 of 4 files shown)


cbeer:
it'd be great to have some updated test docs to play with...

also, i guess we're blocked from deploying until the data is available in -prod?

corylown:
@cbeer I think Jacob is working on updating the our samples docs, which I'll update in the repo once it's available. I guess this is a question for @jacobthill, but since the current counts are wrong, having them appear as zero/blank until the data exists doesn't seem like it should be a blocker.

jacobthill:
@cbeer, @corylown yes I'm trying to get all the data refreshed and new sample data. I've been in meetings or preparing for meetings all day so I haven't been able to work on it much. I'm starting on it now though and tomorrow AM is clear so I should be able to make some progress.

PR #1512, opened 2022-03-01
Add curl + zip for ease of debugging stuff

cbeer:


thatbudakguy (approved):
wow, I just added both these exact things! https://github.com/sul-dlss/dlme/pull/1463/commits/dec7c3d16d0a724960ef793b920fb8c90506d258

Issue #1322, opened 2021-08-31
Make the DLME solr infrastructure more robust.


aaron-collier:
https://app.honeybadger.io/projects/53082/faults/81160819

was the result of SOLR not running on the server. We should have a more robust process for discovering and recovering from this situation..

The current DLME solr instance is running on a single node EC2 / ECS instance for each environment. The webapp node point at the solr node by IP. The solr instance is (seemingly) not directly monitored or alerted on.

Things to consider doing:

- [ ] use DNS names for the solr instance, not IP. This should allow us to build, rebuild and scale the stack without needing to rebuild the webapp environment too
- [ ] expand the production solr instance into a multinode solrcloud cluster
- [ ] adding monitoring + alerting to the solr endpoints
- [ ] (from notes on #1473) the Solr EC2 instance does not (but should) have the Cloudwatch Agent installed so we get disk + RAM alerts from EC2.
- [ ] make sure the solr data is persisted outside the specific node

anarchivist:
A couple points from the most recent Solr outage we saw, raised by @cbeer (feel free to edit if I didn't get the nuance right):

* Solr only has 1 VM associated with it, which means it's not super resilient. Should we add another/set up SolrCloud/etc.?
* Monitoring is a need; Ops didn’t get an alert that the Solr VM’s disk filled up, only that the application was failing its status check
* When we redeploy Solr, we end up blowing the data way in the index - this is less than ideal, so how can we address/improve this?

thatbudakguy:
incident report post-2.4.1 deploy:
1. full reindex was kicked off via ECS exec via the process mentioned in [this comment](https://github.com/sul-dlss/dlme/issues/1476#issuecomment-1048282566)
2. @rsmith11 [reported](https://stanfordlib.slack.com/archives/C5V53K69G/p1646353606210629) that nagios was alerting solr was inaccessible
3. i tried to run a search at dlmenetwork.org and received a 500 error
4. i checked the aws cloudwatch logs and found that the error was due to solr not responding to the spotlight instance
5. @cbeer and i ssh'ed into the bastion host, and then into the solr vm itself. logs in `/var/solr/log/` indicated that the out-of-memory killer had terminated solr
6. `ps aux | grep solr` revealed that solr had been started with the `-Xms512m` option (512meg java heap size), which is the default, but the instance had 4G of RAM available
[...]

PR #1522, opened 2022-03-04
Throttle searches

cbeer:
Part of #1322


thatbudakguy (approved):
merging this so we can test in dev quickly & then hopefully move it to prod

PR #1525, opened 2022-03-04
Update spec/fixtures/ndjson/sample.ndjson

corylown:
Updates the ndjson fixture data to the most recent fields/schema. The updated file has fewer records (90 vs 390) but maybe that's fine?


cbeer:
Having all the expected fields seems like it outweighs the number of records 🤷

PR #1530, opened 2022-03-09
Update fetch s3 invoke to use updated rake task

aaron-collier:
This supports triggering the indexing of resources from airflow in AWS.

### DIFF
--- a/docker/invoke_fetch_s3.sh
+++ b/docker/invoke_fetch_s3.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-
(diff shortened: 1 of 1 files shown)


docker/invoke_fetch_s3.sh
```diff
@@ -3,4 +3,4 @@
```
cbeer, line 3:
```suggestion
# This script is called by the ECS task `dlme_index_from_s3`
# See https://github.com/sul-dlss/terraform-aws/blob/master/organizations/development/dlme/dlme-tasks.tf#L85
```

PR #1548, opened 2022-06-15
Put `skip_pipeline` with the image options, not the url options

cbeer:


app/presenters/dlme_thumbnail_presenter.rb
```diff
@@ -17,6 +17,10 @@ def thumbnail_value(image_options)
     super image_options.reverse_merge(loading: 'lazy')
   end
 
+  def default_thumbnail_value(image_options)
+    super(image_options.except(:skip_pipeline))
+  end
+
```
thatbudakguy, line 23: will this get magically used somewhere without being called? just wondering if that's a blacklight thing

thatbudakguy (approved):

app/presenters/dlme_thumbnail_presenter.rb
cbeer: yes, this just overrides the upstream implementation.

PR #1576, opened 2022-09-23
Remove mysql logic from old db migration that breaks docker build

aaron-collier:
A lot of the cleanup here is just to remove unused code. When building in docker the options parameter throws:

```
dlme-app-1         | wrong number of arguments (given 2, expected 1)
dlme-app-1         | /opt/db/migrate/20170801184720_create_versions.rb:19:in 
```

Removing the options make this migration pass, which makes `versions_table_options` and thus `mysql?` unnecessary. Also, this appears to be a blacklight generated migration, and we use postgres in docker and production, checking for mysql in this app is unecessary.


db/migrate/20170801184720_create_versions.rb
```diff
@@ -1,80 +1,21 @@
...
   def change
-    create_table :versions, versions_table_options do |t|
-      t.string   :item_type, item_type_options
+    create_table :versions do |t|
+      t.string   :item_type, null: false
       t.integer  :item_id,   null: false
       t.string   :event,     null: false
       t.string   :whodunnit
       t.text     :object, limit: TEXT_BYTES
-
```
cbeer, line 25: If we held on to this comment block, I think this migration would match the currently generated upstream one:

https://github.com/paper-trail-gem/paper_trail/blob/master/lib/generators/paper_trail/install/templates/create_versions.rb.erb

aaron-collier: Comment restored :-)

PR #1599, opened 2022-12-05
Switch to shakapacker

jcoyne:
This upgrades webpacker to avoid https://github.com/webpack/webpack/issues/14532


app/javascript/packs/mirador.js
```diff
@@ -15,11 +15,6 @@
 // const images = require.context('../images', true)
 // const imagePath = (name) => images(name, true)
 
-// Polyfills needed for IE and Edge
-import 'core-js/stable';
-import 'url-polyfill/url-polyfill';
-import 'unfetch/polyfill';
-
```
thatbudakguy, line 22: do we no longer need these?

jcoyne: IE is EOL and Edge never needed these.

thatbudakguy: I wondered about Edge; OK to delete them in that case 👍🏻

PR #1617, opened 2023-01-04
Update rubocop rails

jcoyne:
The bug that prevented us from updating it has been resolved


marlo-longley (approved):

marlo-longley:
@jcoyne approved, but needs to be rebased.

PR #1644, opened 2023-02-07
Deploy code locally via Capistrano

### LINKED ISSUE #1610: Capistrano deploy
Deploy the app on prem using capistrano.

jcoyne:
Fixes #1610

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,0 +1,43 @@
+# config valid for current version and patch releases of Capistrano
+lock "~> 3.17.1"
+
+set :application, "dlme"
+set :repo_url, "https://github.com/sul-dlss/dlme.git"
+
+# Default branch is :master
+ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call
+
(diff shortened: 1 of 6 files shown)


config/deploy.rb
```diff
@@ -0,0 +1,43 @@
...
+
+# You can configure the Airbrussh format using :format_options.
+# These are the defaults.
+# set :format_options, command_output: true, log_file: "log/capistrano.log", color: :auto, truncate: :auto
+
+# Default value for :pty is false
+# set :pty, true
+
+# Default value for :linked_files is []
+set :linked_files, %w[config/database.yml]
```
thatbudakguy, line 24: we also linked `config/blacklight.yml` for VT, but that might not matter?

thatbudakguy (approved):

config/deploy.rb
jcoyne: That doesn't make sense unless you're also using shared_configs.

thatbudakguy: Ah, I wondered about that. thx for explaining

PR #1646, opened 2023-02-07
Use the sidekiq redis url variable

jcoyne:


thatbudakguy:
needs rebase since .env is gone

Issue #1607, opened 2023-01-04
On prem monitoring


jcoyne:
Ensure that nagios is pointing at our okcomputer endpoint.

thatbudakguy:
We also want to make sure okcomputer monitors redis, as noted by @corylown

jcoyne:
Blocked until we can deploy to prod. Deploy is blocked by https://github.com/sul-dlss/operations-tasks/issues/3288

jcoyne:
Fixed via https://github.com/sul-dlss/puppet/pull/9133

PR #1660, opened 2023-02-09
Add redis check to okcomputer

thatbudakguy:
Also removes the git version check, which we added for additional
visibility into the built docker images. No longer necessary for
non-docker on-prem deployments.

See #1607

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -2,4 +2,4 @@ OkComputer.mount_at = 'status'
 
 OkComputer::Registry.register 'solr', OkComputer::SolrCheck.new(Blacklight.default_index.connection.uri.to_s.sub(/\/$/, ''))
 
-OkComputer::Registry.register 'version', OkComputer::AppVersionCheck.new(env: 'GIT_INFO')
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -2,4 +2,4 @@
 
 OkComputer::Registry.register 'solr', OkComputer::SolrCheck.new(Blacklight.default_index.connection.uri.to_s.sub(/\/$/, ''))
 
-OkComputer::Registry.register 'version', OkComputer::AppVersionCheck.new(env: 'GIT_INFO')
+OKComputer::Registry.register 'redis', OkComputer::RedisCheck.new(url: ENV.fetch('SIDEKIQ_REDIS_URL', 'redis://localhost:6379/'))
```
jcoyne, line 5: As this is a load balanced system do we want every node to run a check plus the check against the redis server itself? That seems like more checks than necessary.

thatbudakguy: cc @corylown — I think you wanted a redis okcomputer check specifically?

corylown: Since there's a check already in Nagios we probably don't need another one from the app itself.

PR #1672, opened 2023-02-17
Add directions regarding indexing documents

jcoyne:


thatbudakguy (approved):
thanks for adding this to the README!

Issue #1345, opened 2021-11-08
Rename data contributor and item contributor to less confusing names


jacobthill:
The terms ‘data contributor’ and ‘item contributor’ are unclear. Perhaps we can change this to “data provider” and “object steward”, “object holder” or “resource steward”. Also, credit for metadata creation/aggregation/enrichment should be clear as well as credit for those who hold the physical objects.

Facet names can be updated by curator but the same names will need to be implemented in the contributors page.

ggeisler:
It's been a while so I might remember this incorrectly, but based on some notes in a Google doc think Marcin and I discussed this potential solution:

- Rename "Item contributor" to something more generic like “Holding institution”
- Rename "Data contributor" to "Data provider"

And then in the application UI:
- Remove "Data contributor/provider" from facet list
- Continue to display both fields on the item details page (using the new labels)
- Update the section headings on the Statistics page to use the new labels

jacobthill:
I think that's right @ggeisler. I've made you're bullet points into check boxes and implemented the curator configurable changes.

cbeer:
As long as I'm doing this right, it seems ripe for error that the IR field `agg_data_provider` maps to `Holding institution` and `agg_provider` maps to `Data provider`. If we're going with these new labels, can we consider renaming the fields in the IR too?

jacobthill:
@anarchivist what do you think? I feel like I have these names sorted and won't mix them up. The code that maps them never changes. But, as other people start doing data mapping, the odds of making this mistake increase.

anarchivist:
I don't have a strong opinion about what the fields are actually named in the IR, but I will note that the IR field names are intended to map to the EDM properties they're associated with.

jacobthill:
@cbeer, @anarchivist, I think we decided that there is not need to update the field names in the application profile. They are consistent with EDM and I think we can avoid confusion on the data mapping side. I think we have the fields mapped correctly, though Europeana could have chose less confusing names. `edm:Provider` (our `agg_provider`) is "The name or identifier of the organization who delivers data directly to an aggregation service (e.g. Europeana)" while `edm:DataProvider` (our `agg_data_provider`) is "The name or identifier of the organization who contributes data indirectly to an aggregation service (e.g. Europeana)." So, I think we have everything correct and should be okay to call this done and consult the EDM definition every time we do work on the application.

PR #1673, opened 2023-02-17
Highlight item provider on the display page

### LINKED ISSUE #1386: Highlight item providers on item details page
There's some background discussion about this in #1380, but that is probably not required reading to work on this ticket. As noted there, this design does assume that there is **not** a technical constraint that would prevent us from pulling two metadata fields and values from the list of returned item metadata and displaying them outside the `<dl class="document-metadata dl-invert row">` element the currently holds all metadata elements.

Our goal is to better highlight the item and data providers for a given item on its item detail page. The idea is to pull three pieces of metadata already on the item details page and co-locate those in a box that is positioned at the top of the list of metadata fields:

148857352-56b2697a-a600-4ca4-80a5-8788e63d3e24

### Relevant metadata elements

Current item detail page:

Screen Shot 2022-01-12 at 3 25 44 PM

- A: Item contributor (to become `Holding institution` with #1345)
- B: Data contributor (to become `Data provider` with #1345)
- C: Link to the data contributor website

### New contributor highlight box

Screen Shot 2022-01-12 at 3 43 09 PM

Box styling:
- Box should extend as wide as possible to contain the text without line-wrapping (i.e., ideally, each of the three elements in the box display on a single line, though that won't always be possible). But the box shouldn't be fixed-width, because we don't want it to be wider than necessary (and we don't want to have to manually deal with all the different breakpoint width constraints).
- `background-color: #ffffff;`
- `border: 1px solid $color-gray-medium;`
- `border-left: 3px solid $color-black-light;`
- `box-shadow: 0px 4px 4px 0px rgba(0, 0, 0, 0.25);`

Box element styling (as a starting point; we can customize after seeing how this looks):
- The item contributor label becomes `This item resides at:`
- The data contributor label becomes `Item data provided by:`
- Those labels use the default body text styling (`20px, 400 weight, $gray-dark`)
- The values for those labels are links (to the facet value for the respective facet) and should be styled as links
- The "View on contributor website" link is styled as a button horizontally centered in the box. It's a `btn-primary` and I'd try using the `btn-sm` size to keep the box from being taller than necessary.
- The label on the button becomes `View on data provider website`.
- Although visually we want this to look like a button, I think semantically it should remain a link, since it is taking the user outside of the current site?

### Highlight box location
We have two types of item detail page, depending on whether the item is a IIIF item (and thus is displayed in the Mirador viewer) or not. In both cases, the idea is to display the highlight box just above the list of metadata elements. The difference is just that in the IIIF item case we have a wider column in which to display the box:

Screen Shot 2022-01-12 at 4 32 51 PM

I'd suggest horizontally centering the box in the column space available. I'm guessing that'll look the best with boxes of varying widths. 

### Responsive considerations
On narrower viewports, when one of the provider names is long, we'll probably have some line-wrapping. For clarity, I think we should try to wrap the provider name in its own container, rather than treating the whole line as a single container and wrapping from the left-edge. In other words, wrapping would look like this:

Screen Shot 2022-01-12 at 4 33 25 PM

This isn't great (and could increase the box height in certain cases) but hopefully this is a worst case (narrow phone width, long-ish provider title) and won't come up that often. We can look at the wrapping options more closely after it's implemented and adjust if it seems necessary.

jcoyne:
Fixes #1386

### DIFF
--- a/app/components/document_metadata_component.html.erb
+++ b/app/components/document_metadata_component.html.erb
@@ -1,0 +1,8 @@
+<div class="row">
+  <div class="item-provider col-lg-8 offset-lg-2">
+    <%= render(Blacklight::DocumentMetadataComponent.new(fields: @local_fields, show: true)) %>
+    <%= render 'catalog/contributor_link', document: @local_fields.first.document %>
+  </div>
+</div>
+
+<%= render(Blacklight::DocumentMetadataComponent.new(fields: @fields, show: true)) %>
--- a/app/assets/stylesheets/dlme.scss
+++ b/app/assets/stylesheets/dlme.scss
@@ -322,10 +322,6 @@ body {
   padding-bottom: 0.5rem;
 }
 
-.view-on-contributor-website {
-  margin-bottom: 2rem;
-}
-
 .record-feedback {
   position: fixed;
   right: 0;
@@ -377,4 +373,17 @@ body {
 // Don't show multiple title lines in breadcrumb
 .breadcrumb-item > span > :not(:first-child) {
   display: none;
+}
+
+
+.item-provider {
+  background-color: $white;
+  border: 1px solid $color-gray-medium;
+  border-left: 3px solid $color-black-light;
+  box-shadow: 0px 4px 4px 0px rgba(0, 0, 0, 0.25);
+  margin-bottom: 1.5 * $spacer;
+  padding: $spacer;
+  dd, dl {
+    margin-bottom: 0.25rem;
+  }
 }
--- a/app/components/document_metadata_component.rb
+++ b/app/components/document_metadata_component.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Displays the metadata for an item.
+class DocumentMetadataComponent < ViewComponent::Base
+  FIELDS = %w[holding_institution source_institution].freeze
+  def initialize(fields:)
+    super
+    (@local_fields, @fields) = fields.partition { |field| FIELDS.include?(field.key) }
+  end
+end
--- a/app/components/owner_field_component.rb
+++ b/app/components/owner_field_component.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Set our custom layout for the owner field
+class OwnerFieldComponent < Blacklight::MetadataFieldComponent
+  def initialize(layout: OwnerLayoutComponent, **)
+    super
+  end
+end
--- a/app/components/owner_layout_component.rb
+++ b/app/components/owner_layout_component.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Use custom HTML classes for a 4/8 column split
+class OwnerLayoutComponent < Blacklight::MetadataFieldLayoutComponent
+  def initialize(field:, label_class: 'col-md-4', value_class: 'col-md-8')
+    super
+  end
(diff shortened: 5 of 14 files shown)


app/components/document_metadata_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="row">
+  <div class="item-provider col-lg-8 offset-lg-2">
```
thatbudakguy, line 2: Screenshot 2023-02-21 at 12 05 41

this ended up looking cramped for me on most decent-size screens. i think `col-lg-10 offset-lg-1` looked better:

Screenshot 2023-02-21 at 12 06 22

thatbudakguy (commented):
This looks good, but I'm confused about the language shown in Gary's designs ("data provider" and "holding institution"). It seems from https://github.com/sul-dlss/dlme/issues/1345 that the rename should already have happened, but the fields don't appear that way in this PR.

Should this PR be trying to make sure the correct field names display, or did we ultimately decide not to do that? The discussion on that issue is a little unclear. @jacobthill do you know?

app/components/document_metadata_component.html.erb
jcoyne: That seems unbalanced to me.  Almost all of the text is left of center, leaving a bunch of white space on the right.

corylown: @thatbudakguy the field labels come from whatever is configured in the exhibit dashboard, and in prod they've been updated to: "Holding institution" and "Data provider." So I don't think there's an issue with the labels and this PR.

thatbudakguy: Ah, so it's user-defined. That makes sense

thatbudakguy: Re: the sizing of the box, could we just decrease the padding or something? The wrapping that's happening now seemed to be what Gary was explicitly trying to avoid

jcoyne: It's not padding, it's how Bootstrap defines balanced definition lists, which is the proper semantic markup.

thatbudakguy: I'm talking about the padding of the box. In Gary's design it looks smaller than in the PR, which gives the content more space. His design is more similar to the second screenshot I posted.

I don't think we should ship something where the text is wrapping on common screen sizes.

149239215-6b0cdccc-b260-41af-8704-ce3f46972e84

jcoyne: @thatbudakguy I don't think definition lists work that way in bootstrap: https://getbootstrap.com/docs/4.0/content/typography/#description-list-alignment

thatbudakguy: Here's what it looks like with the padding on `.item-provider` (which is our style) decreased to 0.5rem, and switching the column split to 4/8 instead of 3/9 so that the `<dt>` can fit. That way, if text wraps, it's the `<dd>` that are wrapping. I think this is closer to Gary's design.

Screenshot 2023-02-22 at 11 56 06

jcoyne:
@thatbudakguy okay, I've added custom field and layout components for those fields so we can inject that class override.

Screenshot 2023-02-23 at 10 30 14 AM

thatbudakguy:
looks great! thank you

thatbudakguy (approved):

Issue #1372, opened 2021-12-16
Hide the "Sign in" link in the site brandbar


ggeisler:
This could be low priority but it also might be relatively easy and would improve the user experience on the (admittedly rare) occasions when a user selects the "Sign in" link the top-right of the site brandbar.

There are several improvements that could be made when selecting the "Sign in" link:

- User accounts are not applicable to regular DLME users, but we don't make that clear
- We don't display the site masthead on the Sign in / Sign up pages so those pages look sort of broken
- We use inconsistent terminology on these pages ("Sign in" and "Log in")

### Current "sign in" page ("sign up" is basically the same)
<img width="1153" alt="Screen Shot 2021-12-15 at 5 03 34 PM" src="https://user-images.githubusercontent.com/101482/146285493-9e95753a-bac5-4d76-84ba-833d55770182.png">

---

### Suggested updates
- **A:** Add the DLME masthead to the page
- **B:** Add a `<div class="alert alert-warning" role="alert">` box so the user understands this isn't something they should waste time trying. Suggested text is `The Digital Library of the Middle East website does not currently offer user accounts. This page is for website staff only.`
- **C:** Change the "Log in" heading to "Sign in" so it matches the link label

<img width="1098" alt="Screen Shot 2021-12-15 at 5 28 18 PM" src="https://user-images.githubusercontent.com/101482/146286027-b15f9a1d-5bec-40c1-ac4e-58e3df48bacd.png">

---

If we do the above, we should also update the "Sign up" page, since the changes could be pretty much identical (add masthead, add alert box with the same text):

<img width="1098" alt="Screen Shot 2021-12-15 at 5 33 08 PM" src="https://user-images.githubusercontent.com/101482/146286399-16d5e68a-8b80-455c-84c8-f767f1b68fba.png">

mwerla:
@ggeisler I think that if the login/sign up option is not applicable to end users, we should remove that from the public UI. Those few who use that function can simply bookmark the link.

ggeisler:
@mwerla Sure, hiding the Sign in link would be even better. I was concerned that the team would not like the idea of having to bookmark/keep track of the Sign in URL, but if we're okay with that, just hiding the link would be simpler than updating the sign in and sign up pages.

If we go this route, we should make sure to use a technique that hides the Sign link both visually **and** from screen readers, such as adding a `hidden` attribute to the Sign in element (or I think styling it with `display: none` would be functionally equivalent).

[...]

PR #1674, opened 2023-02-17
Navbar UI updates

thatbudakguy:
- Refactor templates to keep Spotlight's header navbar
- Hide the sign in link in the navbar

Ref #1372. First commit makes it so that the navbar still appears on auth pages.


app/assets/stylesheets/dlme.scss
```diff
@@ -169,7 +169,11 @@ body {
 }
 
 .masthead .site-title {
-  white-space: normal;
+  visibility: hidden;
+
+  + small {
```
jcoyne, line 174: What is the purpose of this?  What content is in the small that we want hidden?  Can we just put the hidden style on that element in the template so it's easier to understand the connection?

app/assets/stylesheets/dlme.scss
```diff
@@ -202,6 +206,11 @@ body {
   }
 }
 
+.image-masthead .background-container,
+.image-masthead .background-container-gradient {
+  top: 0;
```
jcoyne, line 211: What is the purpose of this?

app/assets/stylesheets/dlme.scss
```diff
@@ -253,28 +262,34 @@ body {
 }
 
 .topbar {
+  position: absolute;
+  width: 100vw;
```
jcoyne, line 266: Why is these necessary? Topbar is already full width.

jcoyne (changes requested):

app/assets/stylesheets/dlme.scss
thatbudakguy: Spotlight by default inserts the exhibit title and some other content, but we don't want to show it because DLME has only one exhibit and the title is rendered elsewhere. I think this rendering happens in spotlight, so we'd need to override an entire template to do that. Part of the point of this PR was reducing the amount of things we override.

app/assets/stylesheets/dlme.scss
thatbudakguy: The existing styles don't have an effect once its position is changed to `absolute`, which is necessary to overlay it on the masthead.

app/assets/stylesheets/dlme.scss
thatbudakguy: Tested without it and it seemed to work, so I've removed this.

app/assets/stylesheets/dlme.scss
thatbudakguy: Added a comment to note the above

app/assets/stylesheets/dlme.scss
jcoyne: Why are we changing to absolute? I'm really struggling to understand that purpose of these changes.

thatbudakguy: We overrode several spotlight templates in order to construct an HTML setup where the navbar was inside the masthead, in order to implement the current design. This PR gets rid of all the overrides in favor of creating that appearance via a small amount of CSS instead, so we don't diverge from Spotlight so much. It also means that we get to keep the default navbar, so it automatically appears on pages that do not have a masthead, like login pages, which is part of the ticket.

thatbudakguy: Specifically, we use absolute so that the navbar can be positioned over the top of the masthead, which is its following sibling.

app/assets/stylesheets/dlme.scss
```diff
@@ -386,4 +400,13 @@ body {
   dd, dl {
     margin-bottom: 0.25rem;
   }
-}
\ No newline at end of file
+}
+
+// Turn off transparency for the topbar on pages where there's no masthead
```
jcoyne, line 405: Can you add  "(e.g. the pages provided by Devise)"

app/assets/stylesheets/dlme.scss
```diff
@@ -253,17 +258,18 @@ body {
...
-  @include media-breakpoint-down('sm') {
-    margin-bottom: 0;
+  // extra spacing when there's no masthead in between navbar and content
+  + #main-container {
+    padding-top: 8rem !important;
   }
 }
 
-.masthead .topbar .navbar-nav {
+.topbar  .navbar-nav {
```
jcoyne, line 272: Remove extra space

app/assets/stylesheets/dlme.scss
```diff
@@ -340,8 +351,12 @@ body {
     border: 0;
   }
 
-  .masthead .site-title {
-    text-align: center;
+  .masthead .site-title-container {
+    display: none;
```
jcoyne, line 355: Can you add a comma here with the reason we're not displaying it?

app/assets/stylesheets/dlme.scss
```diff
@@ -275,6 +281,11 @@ body {
       text-decoration: underline;
     }
   }
+
+  // hide the "sign in" link
+  &[href="/users/sign_in"] {
```
jcoyne, line 286: Can we remove this and just not render the link https://github.com/sul-dlss/dlme/blob/main/app/views/shared/_user_util_links.html.erb#L41-L45

jcoyne (changes requested):

app/assets/stylesheets/dlme.scss
thatbudakguy: added

app/assets/stylesheets/dlme.scss
thatbudakguy: fixed

app/assets/stylesheets/dlme.scss
thatbudakguy: added

app/assets/stylesheets/dlme.scss
thatbudakguy: Yeah, good call. made that change

jcoyne (approved):

PR #1682, opened 2023-02-21
Suppress the Contact and Sign-in links at narrow widths

### LINKED ISSUE #1388: Hide the "Contact us" link on mobile viewports
Some background discussion of options is in #1384; this ticket is just to make clear the actionable goal.

Depends on #1372 also being done.

The problem we're addressing was stated by @mwerla as:

> One more issue which I think still remains (potentially) confusing is the fact that we have a hamburger icon in the top right corner, and just below it, there is the main menu of the page, which (to me) looks like the expanded menu from the top right hamburger. I am afraid that users may not see the hamburger as yet one more menu to expand.

There are currently three elements in the "utility links" area on the right-side of the brand bar:

- Locale/language dropdown menu
- Sign in link
- Contact us link (which exposes the modal feedback form)

Screen Shot 2022-01-13 at 1 11 24 PM

The three elements are collapsed and hidden behind the hamburger icon toggle at the `sm` viewport and narrower:

Screen Shot 2022-01-13 at 1 11 02 PM

In #1372 we suggest hiding the site "Sign in" link from the brand bar entirely (DLME staff will need to bookmark or remember the URL for the Sign in page, but there are only a handful of us, while there are many, many DLME users to whom the "Sign in" link is useless). So after doing that there will only two elements left in the utility links area. 

The locale/language dropdown menu is essential (users always need to be able to switch to their preferred locale). The "Contact us" link is important, but arguably not essential to be always available from the brand bar location. If we eliminate that link from the utility area -- at the `sm` breakpoint and below only -- we are left with only the locale dropdown to display in the utility area on mobile viewports. 

It appears that there is sufficient horizontal space to display both the site title and the local dropdown menu in the brand bar on mobile viewports, so there would be no need to collapse the utility links and show a hamburger menu; we can just display the locale menu element normally.

So...

- [ ] At the `sm` breakpoint and below, hide the "Contact us" brand bar link
- [ ] If necessary at the `sm` breakpoint, hide the DLME logo image to provide more horizontal space (we already do this at below `sm`)
- [ ] Make any related adjustments necessary to display the locale dropdown menu and eliminating the current expand/collapse section icon and behavior

The result should look similar to this:

148862858-81152ff8-55ac-43af-ab1b-0d43e401b72f

jcoyne:
Fixes #1388
Screenshot 2023-02-21 at 12 07 14 PM
Screenshot 2023-02-21 at 12 07 05 PM
Screenshot 2023-02-21 at 12 06 58 PM

### DIFF
--- a/app/assets/stylesheets/dlme.scss
+++ b/app/assets/stylesheets/dlme.scss
@@ -263,17 +263,17 @@ body {
   }
 }
 
-.masthead .topbar  .navbar-nav {
+.masthead .topbar .navbar-nav {
   text-transform: none;
-}
+  .nav-link {
+    color: $color-gray-lightest;
+    white-space: nowrap;
 
-.masthead .topbar .navbar-nav .nav-link {
-  color: $color-gray-lightest;
-
-  &:active,
-  &:hover,
-  &:focus {
-    text-decoration: underline;
+    &:active,
+    &:hover,
+    &:focus {
+      text-decoration: underline;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


hudajkhan:
I tried this branch out on my local machine, and when making the window width narrower, the contact us and sign in links are suppressed, leaving the locale link. This appears to work as intended.

hudajkhan (approved):

PR #1684, opened 2023-02-21
Implement a simple API endpoint for initiating harvests

### LINKED ISSUE #1683: Enable remote clients to initiate harvesting
Currently the `HarvestsController`:
1. validates you are an authenticated user
2. validates you submitted the form with a CSRF token

In order to initiate a request from Airflow, we need to be able to authenticate a different way (probably a token) and we need to not validate CSRF tokens (as well as make sure we are allowed through CORS).

One way to do this might be to set up a new `/api` route that remote clients can access for doing things like kicking off a harvest. We should secure it appropriately.

thatbudakguy:
Closes #1683

### DIFF
--- a/app/controllers/api/harvests_controller.rb
+++ b/app/controllers/api/harvests_controller.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module Api
+  # Initiate a harvest via the API (e.g. from Airflow)
+  class HarvestsController < ApiController
+    delegate :t, to: :I18n
+
+    def create
+      if request.headers['Authorization'] != ENV.fetch('API_SECRET', '')
+        return render json: { error: t('api.harvest.unauthorized') }, status: :unauthorized
+      end
+
+      return render json: { error: t('api.harvest.no_url') }, status: :bad_request if params['url'].blank?
+
+      return render json: { error: t('api.harvest.no_file') }, status: :bad_request unless File.exist?(filepath)
+
+      return render json: { error: t('api.harvest.duplicate_ids') }, status: :unprocessable_entity if any_duplicate_identifiers?
+
+      AddResourcesJob.perform_later filepath, exhibit: current_exhibit, local: true
+
+      render json: { message: t('api.harvest.success') }, status: :accepted
+    end
+
+    private
+
+    def any_duplicate_identifiers?
+      NdjsonNormalizer.new(body, filename).any_duplicate_identifiers?
+    end
+
+    def body
+      File.read(filepath)
+    end
+
+    def filepath
+      File.join(Settings.data_dir, filename)
+    end
+
+    def filename
+      ActiveStorage::Filename.new(params['url']).sanitized
+    end
+
+    # DLME only has one exhibit
+    def current_exhibit
+      @current_exhibit ||= Spotlight::Exhibit.first
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


app/controllers/api/harvests_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Api
+  # Initiate a harvest via the API (e.g. from Airflow)
+  class HarvestsController < ApiController
+    delegate :t, to: :I18n
+
+    include NdjsonUploadConcern
+
+    rescue_from ActionController::ParameterMissing, with: :no_url
```
jcoyne, line 10: I'm not a huge fan of `rescue_from` because it makes it hard to follow the program flow.  How about wrapping the lines that raise these with `begin ... rescue` instead?

jcoyne (changes requested):

app/controllers/concerns/ndjson_upload_concern.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Shared controller methods for uploading ndjson files for harvesting
+module NdjsonUploadConcern
```
jcoyne, line 4: Can we make this a class instead?  Then we can pass in the required URL... That makes it easier to see what the dependencies of this code is.

app/controllers/api/harvests_controller.rb
thatbudakguy: This was my attempt to placate rubocop's branch condition rule, haha. I'm OK with using `begin`; will change it

app/controllers/concerns/ndjson_upload_concern.rb
thatbudakguy: sure, I like that pattern. will do

thatbudakguy: should it still live in `concerns/`?

jcoyne: No, probably app/services or similar.

app/controllers/api/harvests_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  class HarvestsController < ApiController
+    delegate :t, to: :I18n
+
+    def create
+      begin
+        upload = NdjsonUpload.new(params.require(:url))
+        return invalid_file unless upload.valid?
+      rescue ActionController::ParameterMissing
+        return no_url
+      rescue Errno::ENOENT
```
jcoyne, line 14: Can we move a check for `File.exists?` into `upload.valid?` Then it doesn't have to raise an error in this case?

thatbudakguy: How would we handle the two error cases separately? I think we still want to preserve the distinction between 422 (filepath OK, bad contents) and 400 (filepath not OK)

jcoyne: Have an `error` method on the service object. Once you determine that it's not valid, then you can expect the error method to return a value which can be decoded into a response.

thatbudakguy: I don't think I understand this, but happy to pair if you're willing to explain it

aaron-collier:
Just dropping a comment to say that I've tested this in stage and it's working as expected.

jcoyne (approved):

PR #1696, opened 2023-02-23
Collections for data providers

### LINKED ISSUE #1379: Link collections from each data provider in the contributors page
Depends on https://github.com/sul-dlss/dlme/issues/1327

We would like to expose each collection under the appropriate data provider in the contributors page. 
- [ ] The display value should be whatever value is in the `add_data_provider_collection`.
- [ ] The urls should use the `add_data_provider_collection_id` field since the collection names may change and we don't want the urls to change.
- [ ] We should also display item counts for each collection. Counts can be built from either of the above fields; they should effectively be the same

For the hierarchical type facet we build a field the takes `cho_edm_type` and `cho_has_type` and combines them e.g. if `cho_edm_type` = `Image` and `cho_has_type` = `Photographs` the new field = `['Image', 'Image:Photographs']`. We don't have anything similar for collection but, if needed, @jacobthill can build it.

jcoyne:
Fixes #1379 

Screenshot 2023-03-01 at 3 28 36 PM


corylown (approved):
🎉

Issue #1704, opened 2023-03-01
Trying to "Fetch by URL" with an incorrect file path triggers the "JSON contained duplicate identifiers" error message


jacobthill:
Make sure the duplicate id test accounts for missing files and ideally add a new test to check for incorrect file paths.

corylown:
The validity check on uploaded files can fail for a variety of reasons: https://github.com/sul-dlss/dlme/blob/main/app/services/ndjson_upload.rb#L12

If the validity check finds a problem the error message displayed can be misleading depending on the cause of the failure: https://github.com/sul-dlss/dlme/blob/main/config/locales/en.yml#L51

thatbudakguy:
I think this results from the refactor that I did a bit ago to add the API version of this endpoint, which has some associated error handling. The issue is that the non-API version didn't have any specific error handling other than the duplicate identifier detection, so after the refactor, all errors get treated as that, because there aren't any code paths to throw any other kind of error there.

Probably we just need to borrow https://github.com/sul-dlss/dlme/blob/8b0f2f9307d967659192d9d5c38717dcc93138a0/app/controllers/api/harvests_controller.rb#L22-L49 and adapt to this controller, and also copy the corresponding tests from the API controller and adapt those.

thatbudakguy:
(you could even extract `handle_error` and `current_exhibit` to a concern if you wanted, since that behavior will now be shared between the two controllers)

hudajkhan:
As @thatbudakguy indicated, better handling of the error itself would resolve this situation. Just documenting the current code path below: 
https://github.com/sul-dlss/dlme/blob/main/app/controllers/harvests_controller.rb#L14 shows the call to Ndjsonupload.  Ndjsonupload (https://github.com/sul-dlss/dlme/blob/main/app/services/ndjson_upload.rb#L12)  returns three errors which all match the API controller options above: :no_url, :file_not_found, and :duplicate_ids.  Currently, if the upload method returns any error, the code goes to the invalid_file method https://github.com/sul-dlss/dlme/blob/main/app/controllers/harvests_controller.rb#L24 which uses "dlme_s3s.form.error" as the main go to error (which is the JSON duplicate identifier error).

hudajkhan:
For the moment, not integrating handling for the API and this admin UI harvest code into the same concern.  There may be some refactoring that could be done in that context perhaps separately.

PR #1709, opened 2023-03-02
selecting different error text based on upload error

hudajkhan:


hudajkhan:
Address #1704 .

app/controllers/harvests_controller.rb
```diff
@@ -12,16 +12,29 @@ def create
...
+    error_message = error_message(error)
+    redirect_to spotlight.new_exhibit_resource_path(current_exhibit), flash: { error: error_message }
+  end
+
+  def error_message(error)
+    # NdjsonUpload can also return a ":no_url" error, but the harvest controller requires the url,
+    # so we do not need to check for that situation
+    case error
+    when :duplicate_ids
+      t('dlme_s3s.form.error_duplicate_ids')
```
jcoyne, line 33: Would you mind moving these to `".error.#{error}"`?  It doesn't make sense to talk about `S3` anymore since we're trying to decouple from AWS.

jcoyne (changes requested):

app/controllers/harvests_controller.rb
hudajkhan: I did have a comment about that in the issue and thought maybe that could be a different pull request, but I can make that change here.

hudajkhan: There is a whole dlme_s3s section which may need to go away in general, but maybe I can hold off on that and just move these errors first

hudajkhan: Thanks for walking me through it! Changes in!

jcoyne (approved):

PR #1726, opened 2023-03-03
Upgrade bl-range-limit

jcoyne:
This restores the range limit histogram

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -160,8 +160,9 @@ GEM
       signet
       tophat
       view_component (~> 2.0)
-    blacklight_range_limit (8.2.3)
+    blacklight_range_limit (8.3.0)
       blacklight (>= 7.25.2, < 9)
+      deprecation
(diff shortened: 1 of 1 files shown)


hudajkhan:
Now on Chrome 
Screen Shot 2023-03-03 at 5 07 56 PM

hudajkhan:
Screen Shot 2023-03-03 at 5 09 18 PM
Selected range on Chrome

hudajkhan:
Firefox appears finicky 
Screen Shot 2023-03-03 at 5 11 17 PM
but selecting range then refreshing leads to 
Screen Shot 2023-03-03 at 5 12 06 PM

hudajkhan:
Given that the range limit had disappeared entirely, let's open up a new ticket for fixing what it looks like since it is at least appearing now.

hudajkhan (approved):
Enables the display of the histogram which had previously disappeared.  Still some issues with Firefox display but can be handled separately.

Issue #1572, opened 2022-08-31
Throttling is too aggressive


jacobthill:
I believe throttling was implemented as part of our efforts to improve site performance. I don't know if the settings are different in different environments. As the DLME data manager, I use the development environment for quality checking data transformations. Normal activity for me triggers our throttling which results in a screen that says "Throttled".

This impedes data aggregation work. Can we revisit the original intent behind this and come up with a better solution? Is this implemented in all environments or just dev? Does this have anything to do with bots? If not, humans don't typically use this site except Stanford devs and the DLME data manager.

corylown:
I think it's likely we can close this issue since we're moving the web application on premise, but leaving it open for evaluation during the spring 2023 work cycle.

jcoyne:
Might be rack-attack

thatbudakguy:
deferring this until on-prem migration is mostly complete; then we can test again to see if the situation has changed

jacobthill:
The issue persists after the migration on prem.

thatbudakguy:
moving to "ready"; this needs more analysis to determine the root cause.

thatbudakguy:
@jacobthill can you describe activity that reliably triggers the "throttled" message for you? right now, it looks like the only scenarios we throttle are:
- you try to login more than 20 times in one hour
- you make more than 15 searches in a 15-minute period, where:
  - there is no search term, and
  - you are requesting a page of results deeper than page 10

are these things that you do frequently?

jacobthill:
Its always filtering via facets. When I load a data set like AUC, for example, I need to select each collection then check that the Type facets are correct. So I'm selecting "Data provider" > "Collection id" > "Type", then expanding "Type" for all collections.

thatbudakguy:
got it, thanks for explaining. and do you only do this in stage? or are there scenarios where you would do it in prod as well?

jacobthill:
Only in stage.

thatbudakguy:
confirmed disabled in stage; re-enabled in prod. closing this.

jacobthill:
[...]

PR #1738, opened 2023-03-08
Allow disabling search throttling via config or envvar

thatbudakguy:
Ref #1572

Since this sets the default to disabling search throttling, we can test that it's working correctly and then file a PR in puppet to set `THROTTLE_SEARCHES` to true for prod only to re-enable it in that environment. After that I think the issue can be closed.

### DIFF
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -2,33 +2,36 @@ Rack::Attack.throttle("logins/ip", limit: 20, period: 1.hour) do |req|
   req.ip if req.post? && req.path.start_with?("/users/sign_in")
 end
 
-Rack::Attack.throttle("searches/ip", limit: 15, period: 15.minutes) do |req|
-  # don't throttle requests with a q, because it's more likely to be a real user
-  next if req.params['q'].present?
+# throttle bot-like search queries if configured
+if Settings.throttle_searches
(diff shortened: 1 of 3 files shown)


config/initializers/rack-attack.rb
```diff
@@ -2,33 +2,36 @@
   req.ip if req.post? && req.path.start_with?("/users/sign_in")
 end
 
-Rack::Attack.throttle("searches/ip", limit: 15, period: 15.minutes) do |req|
-  # don't throttle requests with a q, because it's more likely to be a real user
-  next if req.params['q'].present?
+# throttle bot-like search queries if configured
+if ActiveModel::Type::Boolean.new.cast(ENV.fetch("THROTTLE_SEARCHES", Settings.throttle_searches))
```
jcoyne, line 6: How about using `SETTINGS__THROTTLE_SEARCHES` as the env var. Then we can just do `Settings.throttle_searches`. The config gem can cast to boolean and load from ENV vars if properly configured.

jcoyne (changes requested):

config/initializers/rack-attack.rb
thatbudakguy: Testing using the console and this worked exactly how I'd expect...pretty cool. made the change

jcoyne (approved):

PR #1745, opened 2023-03-09
Update i18n-js to version 4

### LINKED ISSUE #1689: Upgrade i18n-js to 4.x
https://github.com/fnando/i18n-js

corylown:
As far as I can tell we're only using the JavaScript components of i18n-js, which as of version 4 are packaged separately from the gem. So I've removed the gem and installed version 4 of i18n-js via yarn.

Closes #1689


jcoyne (changes requested):
Is it possible to keep mirador as the only package compiled by webpack and put i18n in the importmap instead?

corylown:
@jcoyne is there a way to do that and still have i18n-js installed via yarn? I can get the importmap approach to work if I point to cdn supplied i18n-js (which I tried here: https://github.com/sul-dlss/dlme/compare/main...i18n-update). I haven't found a way to use importmap and pin to the yarn installed version. It seemed preferable to install i18n-js via yarn so we get updates via our normal dependency update process.

jcoyne:
> is there a way to do that and still have i18n-js installed via yarn?

No, importmap doesn't use any javascript tooling on the server side or require a javascript runtime.

> It seemed preferable to install i18n-js via yarn so we get updates via our normal dependency update process.

This seems like a pretty big departure from the approach we're already taking with all the rest of the depencencies: https://github.com/sul-dlss/dlme/blob/main/config/importmap.rb

corylown:
Here's the importmap approach: https://github.com/sul-dlss/dlme/pull/1746

PR #1759, opened 2023-03-15
Handle to str error

hudajkhan:
This code handles the error described in https://github.com/orgs/sul-dlss/projects/53?pane=issue&itemId=21913854 .  By distinguishing between an autocomplete request, where we do want strings returned instead of arrays, and other JSON API requests related to the catalog controller where we want arrays returned, we should be able to handle the to_str error which occurs because the autocomplete code in Spotlight was receiving an array instead of the expected string.  Rspec tests have been updated to reflect this difference between a regular JSON API catalog controller request and an autocomplete request.  Added the action name "index" for the regular JSON API catalog controller, since that is one situation (when using the "/raw" option at the end of a catalog page) where the Join render is used again .

### DIFF
--- a/app/processors/join.rb
+++ b/app/processors/join.rb
@@ -23,6 +23,6 @@ class Join < Blacklight::Rendering::AbstractStep
   def json_api_context?
     return false unless context
 
-    context.request&.format&.json? && context.controller.is_a?(CatalogController)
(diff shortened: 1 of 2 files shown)


thatbudakguy (approved):

thatbudakguy:
merging so we can release this today (hopefully!)

hudajkhan:
Thank you!

PR #1760, opened 2023-03-15
Update spotlight to v3.6.0.beta6

thatbudakguy:


yarn.lock
```diff
@@ -4,27 +4,27 @@
 
 "@ampproject/remapping@^2.2.0":
   version "2.2.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
+  resolved "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz"
```
corylown, line 7: Is there a reason why this PR changes all these packages to resolve to npmjs.org instead of yarnpkg.com?

thatbudakguy: because I accidentally used npm instead of yarn! good catch, will fix

thatbudakguy: resolved

corylown (approved):

PR #1877, opened 2024-09-04
Fix flaky tests

corylown:


jcoyne (approved):
These changes seem good, but it would be helpful to understand this a little better. Where was this solution found? What is going on here to make them flaky?

corylown:
When the test that checks "dismiss and don't show again" runs first the other tests fail because the session where "don't show again" was set is still in effect. I tried a bunch of different ways to deal with this based on reviewing methods available in the Capybara driver (https://github.com/teamcapybara/capybara/blob/master/lib/capybara/selenium/driver.rb) and quitting the driver is the only one that worked reliably.

PR #1884, opened 2024-09-20
Add reCAPTCHA to feedback form

taylor-steve:


corylown (changes requested):
@taylor-steve this looks good, except the layout of the "This site is protected..." text in the feedback modal seems off:
Screenshot 2024-11-19 at 2 59 31 PM

Should it be inside the `.modal-body` div and maybe not have the `col-sm-9 offset-sm-3` boostrap classes?

taylor-steve:
@corylown thanks for the catch. I've updated the markup:

Screenshot 2024-11-19 at 12 40 25 PM

corylown (approved):

PR #1887, opened 2024-09-23
Check that search_state.query_param responds to match?

corylown:
Fixes:

```
ActionView::Template::Error (undefined method `match?' for {"id"=>["p15795coll32:68160"]}:ActiveSupport::HashWithIndifferentAccess)
```
which occurs when an item is embedded in a search context, like the exhibit home page.

### DIFF
--- a/spec/models/search_builder_spec.rb
+++ b/spec/models/search_builder_spec.rb
@@ -28,5 +28,13 @@ RSpec.describe SearchBuilder do
         expect(solr_parameters[:mm]).to be_nil
       end
     end
+
+    context "when the query isn't a string" do
+      let(:query) { { q: { a: 1 } } }
+
+      it 'does not set min match' do
+        expect(solr_parameters[:mm]).to be_nil
+      end
+    end
   end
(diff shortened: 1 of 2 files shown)


spec/models/search_builder_spec.rb
```diff
@@ -28,5 +28,13 @@
         expect(solr_parameters[:mm]).to be_nil
       end
     end
+
+    context "when the query isn't a string" do
+      let(:query) { { q: { a: 1 } } }
```
jcoyne, line 33: How does this query occur in the real world?

corylown: When an item is embedded on a page in spotlight.

jcoyne: Is there a page I can go to in order to see this?

jcoyne: Merging this, but we believe this may just be cover for a deeper issue https://github.com/sul-dlss/dlme/issues/1889

PR #1933, opened 2025-01-10
Move oembed to use importmap

jcoyne:


app/assets/config/manifest.js
```diff
@@ -17,4 +17,4 @@
 //= link blacklight-gallery/masonry.js
 //= link blacklight-gallery/slideshow.js
 //= link blacklight-hierarchy/app/assets/javascripts/blacklight/hierarchy/hierarchy.js
-//= link blacklight-oembed/oembed.js
\ No newline at end of file
+//= link blacklight-oembed/oembed.js
```
dnoneill: shouldn't this be removed? Or maybe not? I am a little confused about what this is doing in terms of why this PR is being done.

PR #2056, opened 2026-01-13
Use hosted mirador, get rid of webpacker

jcoyne:


spec/features/mirador_viewer_spec.rb
```diff
@@ -20,12 +20,6 @@
...
   it 'renders a Mirador viewer for IIIF items', js: true do
     visit spotlight.exhibit_solr_document_path(exhibit_id: exhibit.slug, id: '36ebabd9-4d62-4d8e-8e7b-1afd048e872e')
-    expect(page).to have_css('#m3')
-    expect(page).to have_css('.mirador-viewer')
-  end
-
-  it 'enables Mirador Image Tools plugin', js: true do
-    visit spotlight.exhibit_solr_document_path(exhibit_id: exhibit.slug, id: '36ebabd9-4d62-4d8e-8e7b-1afd048e872e')
-    expect(page).to have_css('div[class*=MiradorImageTools]')
+    expect(page).to have_css('iframe[src="https://embed.stanford.edu/iiif?url=https://iiif.bodleian.ox.ac.uk/iiif/manifest/22974622-d838-4496-8835-33ecda85f21f.json"]')
```
dnoneill: I think based on this you need to pass the param for image tools to work.

dnoneill: Oh wait we go rid of that param and it is enabled by default.

jcoyne: Which param is that? They seem to work fine without it:
Screenshot 2026-01-13 at 10 51 27 AM

dnoneill: Yeah I forgot that we decided during one of the work cycles to have image tools available for all Mirador instances. It used to be set by a param but we got rid of that.

PR #2058, opened 2026-01-16
Allow Siteimprove crawlers

### LINKED ISSUE #2057: SiteImprove crawl blocked by robots.txt (and turnstile?)
DLME has a robots.txt that is blocking all crawlers: https://dlmenetwork.org/robots.txt

I'm not sure why we're blocking all crawlers. It would be worth understanding whether this was intentional or not. In either case, we need to ensure siteimprove is not blocked by robots.txt or by turnstile.

Here's the IP and user agent info for siteimprove: https://help.siteimprove.com/support/solutions/articles/80000448553-what-ip-addresses-and-user-agents-are-used-by-siteimprove-

marlo-longley:
Fixes #2057

### DIFF
--- a/config/initializers/bot_challenge_page.rb
+++ b/config/initializers/bot_challenge_page.rb
@@ -29,7 +29,8 @@ Rails.application.config.to_prepare do
   # Also exempt any IPs contained in the CIDR blocks in SAFELIST.
   BotChallengePage::BotChallengePageController.bot_challenge_config.allow_exempt = lambda do |controller, _config|
     (controller.is_a?(CatalogController) && controller.params[:action].in?(%w[facet]) && controller.request.headers['sec-fetch-dest'] == 'empty') ||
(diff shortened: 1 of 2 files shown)


config/initializers/bot_challenge_page.rb
```diff
@@ -32,6 +32,10 @@
       SAFELIST.map { |cidr| IPAddr.new(cidr) }.any? { |range| controller.request.remote_ip.in?(range) }
   end
 
+  BotChallengePage::BotChallengePageController.bot_challenge_config.allow_exempt = ->(controller, config) {
+    controller.request.headers['User-Agent']&.include?('Siteimprove')
```
corylown: I think `controller.request.user_agent` is the preferred way to access the user agent string in Rails. Should this condition be added to the existing `allow_exempt` block above instead of a separate block?

PR #2141, opened 2026-09-15
Shim VIPs for GHSA-xr9x-r78c-5hrm to fix deploy issues

taylor-steve:
(I think this isn't necessary at all if we remove `libvips` from the server after https://github.com/sul-dlss/dlme/pull/2142, waiting to confirm)

Running into failures to deploy, presumably from https://github.com/rails/rails/security/advisories/GHSA-xr9x-r78c-5hrm.

DLME is on Ubuntu 22.04 and cannot easily be updated to Ubuntu 24.04 by Ops at this time, which is what is needed to bring in a compliant version of libvips without compiling from source. This allows us to deploy until that happens. Tested by deploying to stage (and prod...)


taylor-steve:
Confirmed. If `libvips` isn't present at all, disabling the variants is sufficient.

